PR #8, opened 2016-06-10
Issue 3 : I should be able to lauch james-gatling scenario from SBT

chibenwa:
First commit introduced the needed plugin.

Second commit is about reorganization : 
- Split between actual JMAP logic and scenario
- Scenario must be located in test in order to be detected by the plugin

Finaly, I updated documentation about running the given scenario

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,8 +4,11 @@ version := "1.0-SNAPSHOT"
 
 scalaVersion := "2.11.8"
 
+enablePlugins(GatlingPlugin)
+
 libraryDependencies += "io.gatling" % "gatling-core" % "2.2.0"
 libraryDependencies += "io.gatling" % "gatling-http" % "2.2.0"
-
+libraryDependencies += "io.gatling.highcharts" % "gatling-charts-highcharts" % "2.2.0"
+libraryDependencies += "io.gatling" % "gatling-test-framework" % "2.2.0"
 
     
--- a/src/test/scala-2.11/org/apache/james/gatling/jmap/scenari/JmapAuthenticationScenario.scala
+++ b/src/test/scala-2.11/org/apache/james/gatling/jmap/scenari/JmapAuthenticationScenario.scala
@@ -1,7 +1,7 @@
 package org.apache.james.gatling.jmap.scenari
 
 import io.gatling.core.Predef._
-import io.gatling.http.Predef._
+import org.apache.james.gatling.jmap.HttpSettings
 import org.apache.james.gatling.jmap.JmapAuthentication.authentication
 
 class JmapAuthenticationScenario extends Simulation {
@@ -10,14 +10,9 @@ class JmapAuthenticationScenario extends Simulation {
   val username = "username@" + domain
   val password = "password"
 
-  val httpProtocol = http
-    .baseURL("http://127.0.0.1")
-    .acceptHeader("application/json")
-    .contentTypeHeader("application/json; charset=UTF-8")
-
   val scn = scenario("JmapAuthentication")
     .exec(authentication(username, password))
 
-  setUp(scn.inject(atOnceUsers(1))).protocols(httpProtocol)
+  setUp(scn.inject(atOnceUsers(1))).protocols(HttpSettings.httpProtocol)
 
 }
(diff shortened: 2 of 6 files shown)


build.sbt
```diff
@@ -4,8 +4,11 @@ version := "1.0-SNAPSHOT"
 
 scalaVersion := "2.11.8"
 
+enablePlugins(GatlingPlugin)
+
 libraryDependencies += "io.gatling" % "gatling-core" % "2.2.0"
 libraryDependencies += "io.gatling" % "gatling-http" % "2.2.0"
-
+libraryDependencies += "io.gatling.highcharts" % "gatling-charts-highcharts" % "2.2.0"
```
mbaechler, line 11: I'm pretty sure there's a license problem with that

chibenwa, line 11: License can be found here : https://raw.githubusercontent.com/gatling/gatling-highcharts/master/LICENSE . Clearly it is not free.

I'm not a License expert but

I agree you can't use a library licensed with **GNU Affero General Public License version 3** in a less permissive project. Hopefully this is not what I am doing here.

However I don't see what prevents  less permissive librairy to be used by a **GNU Affero General Public License version 3**...

I want also to quote that without this librairy, I can not generate reports : 

```
[error] Uncaught exception when running org.apache.james.gatling.jmap.scenari.JmapAuthenticationScenario: java.lang.NoClassDefFoundError: io/gatling/charts/component/impl/ComponentLibraryImpl
[trace] Stack trace suppressed: run last gatling:test for the full output.
[info] Simulation(s) execution ended.
[error] Error during tests:
[error]     org.apache.james.gatling.jmap.scenari.JmapAuthenticationScenario
[error] (gatling:test) sbt.TestsFailedException: Tests unsuccessful
```

I can not run gatling from sbt without it.

chibenwa, line 11: I am reading the AFFERO license....

gatling.highcharts fall on the definition of a system library : 

> The "System Libraries" of an executable work include anything, other than the
> work as a whole, that (a) is included in the normal form of packaging a Major
> Component, but which is not part of that Major Component, and (b) serves only
> to enable use of the work with that Major Component, or to implement a Standard
> Interface for which an implementation is available to the public in source code
> form. A "Major Component", in this context, means a major essential component
> (kernel, window system, and so on) of the specific operating system (if any) on
> which the executable work runs, or a compiler used to produce the work, or an
> object code interpreter used to run it.

It is stated that : 

> The "Corresponding Source" [...] does not include the work's System Libraries

chibenwa, line 11: > A separable portion of the object code, whose source code is excluded from the
> Corresponding Source as a System Library, need not be included in conveying the
> object code work.

This means **gatling-highcharts** is not covered by the Affero license

chibenwa, line 11: And I don't see which part of this I am breaking by integrating it in this work : https://raw.githubusercontent.com/gatling/gatling-highcharts/master/LICENSE

LSC-LINAGORA, line 11: HighCharts is nonfree. Please don't use it as a dependency.

mbaechler:
This project is a _tool_, not a _product_. We don't expect to ship it to anybody but we will only use it on our own computers. Our code just describe it needs this library to run, but doesn't include the component, I can't see anything wrong in merging this PR.
We will be careful not to distributed this project in other forms, we have to keep it as simple source code.

src/test/scala-2.11/org/apache/james/gatling/jmap/scenari/JmapAuthenticationScenario.scala
```diff
@@ -0,0 +1,23 @@
...
+import io.gatling.http.Predef._
+import org.apache.james.gatling.jmap.JmapAuthentication.authentication
+
+class JmapAuthenticationScenario extends Simulation {
+
+  val domain = "domain-jmapauthentication.tld"
+  val username = "username@" + domain
+  val password = "password"
+
+  val httpProtocol = http
```
aduprat: This may be shared as it will be used everywhere.

chibenwa: Done

aduprat:
:+1:

mbaechler:
:+1:

rouazana:
:+1:

PR #13, opened 2016-06-24
ISSUE-12 Log level should be info

chibenwa:


src/test/resources/logback.xml
```diff
@@ -0,0 +1,12 @@
+<configuration>
```
mbaechler: it should be logback-test.xml actually

chibenwa: Ok

Issue #6, opened 2016-06-09
A user should be able to send a mail to an other user at random


chibenwa:
The mail subject and content should be provisionned.

The mail recipient should be one of the provisionned users.

chibenwa:
Depends on #4

PR #14, opened 2016-06-28
Issue 6 I should be able to send a mail randomly using gatling

chibenwa:
Solves #6

First we should read mailboxes. I wrote a pipeline for reading the JMAP response and add it to the session. I'm 100% Ok to rework this part to propose something simpler.

Then select a user in the feeder and send him a message. UUID are still used to generate random strings.

### DIFF
--- a/src/main/scala-2.11/org/apache/james/gatling/control/JamesWebAdministration.scala
+++ b/src/main/scala-2.11/org/apache/james/gatling/control/JamesWebAdministration.scala
@@ -25,4 +25,7 @@ object JamesWebAdministration {
     wsClient.url(s"$baseUrl/users/${user.username.value}")
       .put(s"""{"password":"${user.password.value}"}""")
 
+  def createInbox(username: Username) = wsClient.url(s"$baseUrl/users/${username.value}/mailboxes/INBOX").put("")
+  def createOutbox(username: Username) = wsClient.url(s"$baseUrl/users/${username.value}/mailboxes/outbox").put("")
+
 }
--- a/src/main/scala-2.11/org/apache/james/gatling/control/JamesWebAdministration.scala
+++ b/src/main/scala-2.11/org/apache/james/gatling/control/JamesWebAdministration.scala
@@ -44,22 +44,22 @@ object JamesWebAdministration {
       .put(s"""{"password":"${user.password.value}"}""")
       .map(response => user)
 
-  def getMailboxesUrl(username: Username): String = {
-    s"$baseUrl/users/${username.value}/mailboxes"
+  def getMailboxesUrl(username: Username): URL = {
+    new URL(s"$baseUrl/users/${username.value}/mailboxes")
   }
 
   def createInbox(username: Username) = {
-    val mailboxesUrl: String = getMailboxesUrl(username)
+    val mailboxesUrl = getMailboxesUrl(username)
     wsClient.url(s"$mailboxesUrl/INBOX").put("")
   }
 
   def createOutbox(username: Username) = {
-    val mailboxesUrl: String = getMailboxesUrl(username)
+    val mailboxesUrl = getMailboxesUrl(username)
     wsClient.url(s"$mailboxesUrl/outbox").put("")
   }
 
   def createSentBox(username: Username) = {
-    val mailboxesUrl: String = getMailboxesUrl(username)
+    val mailboxesUrl = getMailboxesUrl(username)
     wsClient.url(s"$mailboxesUrl/sent").put("")
   }
 
--- a/src/main/scala-2.11/org/apache/james/gatling/control/JamesWebAdministration.scala
+++ b/src/main/scala-2.11/org/apache/james/gatling/control/JamesWebAdministration.scala
@@ -2,9 +2,10 @@ package org.apache.james.gatling.control
 
 import java.net.URL
 
-import play.api.libs.ws.WSResponse
+import org.apache.james.gatling.utils.RandomStringGenerator
 import play.api.libs.ws.ning.NingWSClient
 
+import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 
 case class Domain(value: String)
@@ -12,20 +13,54 @@ case class Username(value: String)
 case class Password(value: String)
 case class User(username: Username, password: Password)
 
+object User {
+  def random(domain: Domain) = {
+    val localAddressPart = RandomStringGenerator.randomString
+    User(Username(s"$localAddressPart@${domain.value}"),
+      Password.random)
+  }
+}
+
+object Domain {
+  def random = Domain(RandomStringGenerator.randomString)
+}
+
+object Password {
+  def random = Password(RandomStringGenerator.randomString)
+}
+
 object JamesWebAdministration {
 
   val wsClient = NingWSClient()
 
   val baseUrl = new URL("http://172.17.0.4:8000")
 
-  def addDomain(domain: Domain): Future[WSResponse] = wsClient.url(s"$baseUrl/domains/${domain.value}")
+  def addDomain(domain: Domain): Future[Domain] = wsClient.url(s"$baseUrl/domains/${domain.value}")
     .put("")
+    .map(response => domain)
 
-  def addUser(user: User): Future[WSResponse] =
+  def addUser(user: User): Future[User] =
     wsClient.url(s"$baseUrl/users/${user.username.value}")
       .put(s"""{"password":"${user.password.value}"}""")
+      .map(response => user)
 
-  def createInbox(username: Username) = wsClient.url(s"$baseUrl/users/${username.value}/mailboxes/INBOX").put("")
-  def createOutbox(username: Username) = wsClient.url(s"$baseUrl/users/${username.value}/mailboxes/outbox").put("")
+  def getMailboxesUrl(username: Username): String = {
+    s"$baseUrl/users/${username.value}/mailboxes"
+  }
+
+  def createInbox(username: Username) = {
+    val mailboxesUrl: String = getMailboxesUrl(username)
+    wsClient.url(s"$mailboxesUrl/INBOX").put("")
+  }
+
+  def createOutbox(username: Username) = {
+    val mailboxesUrl: String = getMailboxesUrl(username)
+    wsClient.url(s"$mailboxesUrl/outbox").put("")
+  }
+
+  def createSentBox(username: Username) = {
+    val mailboxesUrl: String = getMailboxesUrl(username)
+    wsClient.url(s"$mailboxesUrl/sent").put("")
+  }
 
 }
--- a/src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
+++ b/src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
@@ -9,22 +9,52 @@ import scala.concurrent.Future
 
 object UserFeeder {
 
-  def createUserFeeder(userCount: Int): Array[Map[String, String]] = {
-    val domain = Domain(UUID.randomUUID().toString)
+  val USERNAME: String = "username"
+  val PASSWORD: String = "password"
 
+  def createUserFeeder(userCount: Int): Array[Map[String, String]] = {
+    toFeeder(createRegisteredUsers(userCount))
+  }
+
+  def createUserFeederWithInboxAndOutbox(userCount: Int): Array[Map[String, String]] = {
+    toFeeder(
+      decorateWithInboxAndOutbox(
+        createRegisteredUsers(userCount)))
+  }
+
+  def decorateWithInboxAndOutbox(users: List[User]): List[User] = {
+    Future.sequence(
+      users.map(user => Future.sequence(
+        List(JamesWebAdministration.createInbox(user.username),
+          JamesWebAdministration.createOutbox(user.username))))
+    ).get
+    users
+  }
+
+  def createRegisteredUsers(userCount: Int): List[User] = {
+    val domain = Domain(UUID.randomUUID().toString)
     JamesWebAdministration.addDomain(domain).get
 
-    val users = (0 until userCount)
+    registerUsers(generateUsers(userCount, domain))
+  }
+
+  def generateUsers(userCount: Int, domain: Domain): List[User] = {
+    (0 until userCount)
       .map(i => User(
         Username(s"""${UUID.randomUUID().toString}@${domain.value}"""),
         Password(UUID.randomUUID().toString)))
       .toList
+  }
 
+  def registerUsers(users: List[User]): List[User] = {
     Future.sequence(
       users.map(user => JamesWebAdministration.addUser(user)))
       .get
+    users
+  }
 
-    users.map(user => Map("username" -> user.username.value, "password" -> user.password.value)).toArray
+  def toFeeder(users: List[User]): Array[Map[String, String]] = {
+    users.map(user => Map(USERNAME -> user.username.value, PASSWORD -> user.password.value)).toArray
   }
 
 }
--- a/src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
+++ b/src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
@@ -1,60 +1,59 @@
 package org.apache.james.gatling.control
 
-import java.util.UUID
-
 import io.gatling.core.Predef._
 
+import scala.concurrent.duration.Duration.Inf
+import scala.concurrent.{Await, Future}
 import scala.concurrent.ExecutionContext.Implicits.global
-import scala.concurrent.Future
+import scala.util.Success
 
 object UserFeeder {
 
-  val USERNAME: String = "username"
-  val PASSWORD: String = "password"
+  type UserFeeder = Array[Map[String, String]]
 
-  def createUserFeeder(userCount: Int): Array[Map[String, String]] = {
-    toFeeder(createRegisteredUsers(userCount))
-  }
+  val USERNAME_SESSION_PARAM: String = "username"
+  val PASSWORD_SESSION_PARAM: String = "password"
 
-  def createUserFeederWithInboxAndOutbox(userCount: Int): Array[Map[String, String]] = {
+  def createUserFeeder(userCount: Int): UserFeeder =
     toFeeder(
-      decorateWithInboxAndOutbox(
-        createRegisteredUsers(userCount)))
-  }
+      awaitInitialization(
+        createRegisteredUserFutures(userCount)))
 
-  def decorateWithInboxAndOutbox(users: List[User]): List[User] = {
-    Future.sequence(
-      users.map(user => Future.sequence(
-        List(JamesWebAdministration.createInbox(user.username),
-          JamesWebAdministration.createOutbox(user.username))))
-    ).get
-    users
-  }
+  def createUserFeederWithInboxAndOutbox(userCount: Int): UserFeeder =
+    toFeeder(
+      awaitInitialization(
+        createRegisteredUserFutures(userCount)
+          .map(userFuture => userFuture.andThen {
+            case Success(user) => registerSystemMailboxes(user)
+          })))
 
-  def createRegisteredUsers(userCount: Int): List[User] = {
-    val domain = Domain(UUID.randomUUID().toString)
+  private def awaitInitialization(userFutures: List[Future[User]]) =
+    Await.result(
+      Future.sequence(userFutures),
+      Inf)
+
+  private def createRegisteredUserFutures(userCount: Int): List[Future[User]] = {
+    val domain = Domain.random
     JamesWebAdministration.addDomain(domain).get
 
-    registerUsers(generateUsers(userCount, domain))
+    generateUsers(userCount, domain)
+      .map(JamesWebAdministration.addUser)
   }
 
-  def generateUsers(userCount: Int, domain: Domain): List[User] = {
+  private def generateUsers(userCount: Int, domain: Domain): List[User] =
     (0 until userCount)
-      .map(i => User(
-        Username(s"""${UUID.randomUUID().toString}@${domain.value}"""),
-        Password(UUID.randomUUID().toString)))
+      .map(i => User.random(domain))
       .toList
-  }
 
-  def registerUsers(users: List[User]): List[User] = {
+  def registerSystemMailboxes(user: User): Future[User] = {
     Future.sequence(
-      users.map(user => JamesWebAdministration.addUser(user)))
-      .get
-    users
+      List(JamesWebAdministration.createInbox(user.username),
+        JamesWebAdministration.createOutbox(user.username),
+        JamesWebAdministration.createSentBox(user.username)))
+      .map(responseList => user)
   }
 
-  def toFeeder(users: List[User]): Array[Map[String, String]] = {
-    users.map(user => Map(USERNAME -> user.username.value, PASSWORD -> user.password.value)).toArray
-  }
+  private def toFeeder(users: List[User]): UserFeeder =
+    users.map(user => Map(USERNAME_SESSION_PARAM -> user.username.value, PASSWORD_SESSION_PARAM -> user.password.value)).toArray
 
 }
--- a/src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
+++ b/src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
@@ -1,0 +1,31 @@
+package org.apache.james.gatling.jmap
+
+import io.gatling.core.Predef._
+import io.gatling.http.Predef._
+import org.apache.james.gatling.utils.JmapChecks
+
+object JmapMailboxes {
+
+  private val mailboxListPath: String = "[0][1].list"
+
+  def getMailboxes = JmapAuthentication.authenticatedQuery("getMailboxes", "/jmap")
+    .body(StringBody("""[["getMailboxes", {}, "#0"]]"""))
+    .check(status.is(200))
+    .check(JmapChecks.noError)
+
+  def getMailboxIds = {
+    val idPaths = s"$$$mailboxListPath[*].id"
+    getMailboxes
+      .check(jsonPath(idPaths).findAll.saveAs("mailboxIds"))
+  }
+
+  def getSystemMailboxes = {
+    val inboxIdPath = s"$$$mailboxListPath[?(@.role == 'inbox')].id"
+    val outboxIdPath = s"$$$mailboxListPath[?(@.role == 'outbox')].id"
+    val sentIdPath = s"$$$mailboxListPath[?(@.role == 'sent')].id"
+    getMailboxes
+      .check(jsonPath(inboxIdPath).saveAs("inboxMailboxId"))
+      .check(jsonPath(outboxIdPath).saveAs("outboxMailboxId"))
+      .check(jsonPath(sentIdPath).saveAs("sentMailboxId"))
+  }
+}
--- a/src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMessages.scala
+++ b/src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMessages.scala
@@ -1,0 +1,53 @@
+package org.apache.james.gatling.jmap
+
+import java.util.UUID
+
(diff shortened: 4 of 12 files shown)


src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  /* The request to process */
+
+  def allMailboxesRequest = JmapAuthentication.authenticatedQuery("getMailboxes", "/jmap")
+    .body(StringBody("""[["getMailboxes", {}, "#0"]]"""))
+    .check(status.is(200))
+    .check(jsonPath("error").notExists)
+
+  class PipelineElement (val dataExtractor: HttpRequestBuilder => HttpRequestBuilder, val next: Option[PipelineElement]) {
```
mbaechler: `val next: Option[PipelineElement] = None`

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
```diff
@@ -0,0 +1,55 @@
...
+
+    def this(dataExtractor: HttpRequestBuilder => HttpRequestBuilder) = this(dataExtractor, None)
```
mbaechler: and this becomes useless

chibenwa: Ok thanks, I missed this bit of scala syntax

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  def extractMailboxesIds(httpRequestBuilder: HttpRequestBuilder) = httpRequestBuilder
+    .check(jsonPath("$.mailboxes.list[*].id").saveAs("mailboxIds"))
+
+  def extractMailboxWithRole(role: String, sessionVariableName: String)(httpRequestBuilder: HttpRequestBuilder) =
+    extractMailboxWithAttribute("role")(role, sessionVariableName)(httpRequestBuilder)
+  def extractMailboxWithName(name: String, sessionVariableName: String)(httpRequestBuilder: HttpRequestBuilder) =
+    extractMailboxWithAttribute("name")(name, sessionVariableName)(httpRequestBuilder)
+
+  def extractMailboxWithAttribute(attributeKey: String)(value: String, sessionVariableName: String)(httpRequestBuilder: HttpRequestBuilder): HttpRequestBuilder = {
```
mbaechler: I know that curying is fancy but why don't you simply return a HttpCheck ?

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
```diff
@@ -0,0 +1,55 @@
+package org.apache.james.gatling.jmap
+
+import io.gatling.core.Predef._
+import io.gatling.http.Predef._
+
+import io.gatling.http.request.builder.HttpRequestBuilder
+
+object JmapMailboxes {
+
+  def getMailboxes(dataExtractors: (HttpRequestBuilder => HttpRequestBuilder)*) =
```
mbaechler: you really should define a type for `(HttpRequestBuilder => HttpRequestBuilder)`

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
```diff
@@ -0,0 +1,55 @@
...
+      dataExtractors.map(dataExtractor => new PipelineElement(dataExtractor))
+        .reduce(_ combineWith _)
+        .applyPipeline(allMailboxesRequest)
+    } else {
+      allMailboxesRequest
+    }
+
+  /* available data extractors */
+
+  def extractSentId(httpRequestBuilder: HttpRequestBuilder) = extractMailboxWithRole("sent", "sentMailboxId")(httpRequestBuilder)
```
mbaechler: And this should be instances of the type previously defined

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
mbaechler: what is the output type ?

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
```diff
@@ -0,0 +1,55 @@
+package org.apache.james.gatling.jmap
+
+import io.gatling.core.Predef._
+import io.gatling.http.Predef._
+
+import io.gatling.http.request.builder.HttpRequestBuilder
+
+object JmapMailboxes {
```
mbaechler, line 7: the whole solution is really overkill. I think that you could just design your extractors as checks and then chain calls to these checks after the HttpRequest.

src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
```diff
@@ -9,22 +9,52 @@ import scala.concurrent.Future
...
-    val users = (0 until userCount)
+    registerUsers(generateUsers(userCount, domain))
+  }
+
+  def generateUsers(userCount: Int, domain: Domain): List[User] = {
+    (0 until userCount)
       .map(i => User(
         Username(s"""${UUID.randomUUID().toString}@${domain.value}"""),
         Password(UUID.randomUUID().toString)))
       .toList
```
mbaechler, line 46: why do you need to materialize your list ?

src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
```diff
@@ -9,22 +9,52 @@ import scala.concurrent.Future
...
+    ).get
+    users
+  }
+
+  def createRegisteredUsers(userCount: Int): List[User] = {
+    val domain = Domain(UUID.randomUUID().toString)
     JamesWebAdministration.addDomain(domain).get
 
-    val users = (0 until userCount)
+    registerUsers(generateUsers(userCount, domain))
```
mbaechler: the natural way to write this is more like : 

```
users = (0 until userCount)
  .map(generateUser(domain))
  .map(registerUser(_))
waitFor(users)
```

src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
```diff
@@ -9,22 +9,52 @@ import scala.concurrent.Future
...
+    toFeeder(createRegisteredUsers(userCount))
+  }
 
+  def createUserFeederWithInboxAndOutbox(userCount: Int): Array[Map[String, String]] = {
+    toFeeder(
+      decorateWithInboxAndOutbox(
+        createRegisteredUsers(userCount)))
+  }
+
+  def decorateWithInboxAndOutbox(users: List[User]): List[User] = {
```
mbaechler: `map()` is composable, I think you can do the same thing without decorators

src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
```diff
@@ -9,22 +9,52 @@ import scala.concurrent.Future
...
     JamesWebAdministration.addDomain(domain).get
 
-    val users = (0 until userCount)
+    registerUsers(generateUsers(userCount, domain))
+  }
+
+  def generateUsers(userCount: Int, domain: Domain): List[User] = {
+    (0 until userCount)
       .map(i => User(
         Username(s"""${UUID.randomUUID().toString}@${domain.value}"""),
```
mbaechler: come on, you really wrote code that needs execution into a string ?

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMessages.scala
```diff
@@ -0,0 +1,52 @@
...
+              }
+            }
+          },
+          "#0"
+          ]]"""))
+      .check(status.is(200))
+      .check(jsonPath("$.error").notExists)
+
+  def sendMessagesRandomly(feeder: Array[Map[String, String]]) =
+    sendMessages(
```
mbaechler, line 41: please use named arguments here for clarity

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMessages.scala
```diff
@@ -0,0 +1,52 @@
...
+      MessageId(UUID.randomUUID().toString),
+      selectectRecipientAtRandom(feeder),
+      Subject(UUID.randomUUID().toString),
+      TextBody(UUID.randomUUID().toString))
+
+  def selectectRecipientAtRandom(feeder: Array[Map[String, String]]) =
+    RecipientAddress(
+      feeder(Math.abs(Random.nextInt() % feeder.length))
```
mbaechler: `Random.nextInt(feeder.size)`

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMessages.scala
```diff
@@ -0,0 +1,52 @@
...
+          ]]"""))
+      .check(status.is(200))
+      .check(jsonPath("$.error").notExists)
+
+  def sendMessagesRandomly(feeder: Array[Map[String, String]]) =
+    sendMessages(
+      MessageId(UUID.randomUUID().toString),
+      selectectRecipientAtRandom(feeder),
+      Subject(UUID.randomUUID().toString),
+      TextBody(UUID.randomUUID().toString))
```
mbaechler: I think it's time to stop repeating this awful UUID generation call.
Please define a RandomContentFactory (that wraps this call)

src/test/scala-2.11/org/apache/james/gatling/jmap/scenari/JmapSendMessagesScenario.scala
```diff
@@ -0,0 +1,30 @@
...
+  val feeder = UserFeeder.createUserFeederWithInboxAndOutbox(userCount)
+
+  val scn = scenario("JmapSendMessages")
+    .feed(feeder)
+    .pause(1 second, 10 second)
+    .exec(authentication())
+    .pause(1 second)
+    .exec(JmapMailboxes.getMailboxes(JmapMailboxes.extractInboxId, JmapMailboxes.extractOutboxId))
+    .pause(1 second)
+    .repeat(360, "any") {
```
mbaechler: why "any" ?

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
chibenwa, line 7: Reworked

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMessages.scala
chibenwa, line 41: I improved readability by adding default argument for the case class here.
You just have the argument creation threw constructor, which have the benefit of both being readable and succint.

src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
chibenwa, line 46: Because I prefer a List rather than a TraverseableLike as an argument of other methods, and a traverseable like is not a List :-/

src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
mbaechler: no answer about that ?

src/main/scala-2.11/org/apache/james/gatling/control/JamesWebAdministration.scala
```diff
@@ -2,14 +2,15 @@ package org.apache.james.gatling.control
...
 import java.net.URL
 
+import org.apache.james.gatling.utils.RandomStringGenerator
 import play.api.libs.ws.WSResponse
 import play.api.libs.ws.ning.NingWSClient
 
 import scala.concurrent.Future
 
-case class Domain(value: String)
+case class Domain(value: String = RandomStringGenerator.randomString)
```
mbaechler: generating a random value for domain (and other classes) doesn't sound like an expected default behavior.
You should have a factory with a named method like `Domain.generateRandom()` (Domain being an object in that case)

src/main/scala-2.11/org/apache/james/gatling/control/JamesWebAdministration.scala
```diff
@@ -25,4 +26,8 @@ object JamesWebAdministration {
     wsClient.url(s"$baseUrl/users/${user.username.value}")
       .put(s"""{"password":"${user.password.value}"}""")
 
+  def createInbox(username: Username) = wsClient.url(s"$baseUrl/users/${username.value}/mailboxes/INBOX").put("")
```
mbaechler: so many repetition of the same paths

src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
```diff
@@ -1,30 +1,68 @@
...
-import scala.concurrent.Future
+import scala.concurrent.duration.Duration.Inf
+import scala.concurrent.{Await, Future}
+import scala.util.Success
 
 object UserFeeder {
 
-  def createUserFeeder(userCount: Int): Array[Map[String, String]] = {
-    val domain = Domain(UUID.randomUUID().toString)
+  val USERNAME: String = "username"
```
mbaechler: I guess it's not really a username (:

src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
```diff
@@ -1,30 +1,68 @@
...
+  val PASSWORD: String = "password"
+
+  def createUserFeeder(userCount: Int): Array[Map[String, String]] =
```
mbaechler: `Array[Map[String, String]]` is not an helpful type, you have type aliases in Scala.
`type UserFeeder = Array[Map[String, String]]`

src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
```diff
@@ -1,30 +1,68 @@
...
+          .map(addSystemMailboxes(_))))
 
+  private def awaitInitialization(userFutures: List[Future[User]]) =
+    Await.result(
+      Future.sequence(userFutures),
+      Inf)
+
+  private def createRegisteredUserFutures(userCount: Int): List[Future[User]] = {
+    val domain = Domain()
     JamesWebAdministration.addDomain(domain).get
```
mbaechler, line 37: does this work : `addDomain(domain).map(generateUsers(userCount, _).map(registerUser(_)))` ?

src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
```diff
@@ -1,30 +1,68 @@
...
+
+  private def generateUsers(userCount: Int, domain: Domain): List[User] =
+    (0 until userCount)
+      .map(i => createRandomUser(domain))
       .toList
 
-    Future.sequence(
-      users.map(user => JamesWebAdministration.addUser(user)))
-      .get
+  private def createRandomUser(domain: Domain): User = {
```
mbaechler: It could be a User factory method

src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
```diff
@@ -1,30 +1,68 @@
...
+  }
 
-    users.map(user => Map("username" -> user.username.value, "password" -> user.password.value)).toArray
+  private def registerUser(user: User): Future[User] = {
+      JamesWebAdministration.addUser(user)
+        .map(response => user)
   }
 
+  private def addSystemMailboxes(userFuture: Future[User]): Future[User] =
+    userFuture.andThen {
```
mbaechler: instead of chaining the call, you should return the function to apply and do the chaining at an upper level

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
mbaechler, line 7: It's still not what I meant : 

```
def extractMailboxIds(key: String) : List[Check] = List(jsonPath("$[0][1].list[*].id").findAll.saveAs(key))
```

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
```diff
@@ -0,0 +1,22 @@
...
+  def getMailboxes = JmapAuthentication.authenticatedQuery("getMailboxes", "/jmap")
+    .body(StringBody("""[["getMailboxes", {}, "#0"]]"""))
+    .check(status.is(200))
+    .check(jsonPath("$.error").notExists)
+
+  def getMailboxIds = getMailboxes
+    .check(jsonPath("$[0][1].list[*].id").findAll.saveAs("mailboxIds"))
+
+  def getSystemMailboxes = getMailboxes
+    .check(jsonPath("$[0][1].list[?(@.role == 'inbox')].id").saveAs("inboxMailboxId"))
```
mbaechler: what is `$[0][1].list` ? repeated so many times

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMessages.scala
```diff
@@ -0,0 +1,53 @@
...
+                "to":  [{"name":"${recipientAddress.address}", "email": "${recipientAddress.address}"}],
+                "textBody": "${textBody.text}",
+                "subject": "${subject.subject}",
+                "mailboxIds": ["$${outboxMailboxId}"]
+              }
+            }
+          },
+          "#0"
+          ]]"""))
+      .check(status.is(200))
```
mbaechler, line 37: these two checks are so common that they should be in a JmapChecks object

src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
chibenwa: I added a map clause, and modifies the underlying future with a andThen clause.

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMessages.scala
mbaechler, line 41: I still think using named argument during call would be a gain : 

```
sendMessages(
  messageId = MessageId.generate,
  to = selectectRecipientAtRandom(feeder),
  subject = Subject.generateRandom,
  textbody = TextBody.generateRandom)
```

src/test/scala-2.11/org/apache/james/gatling/jmap/scenari/JmapSendMessagesScenario.scala
```diff
@@ -0,0 +1,30 @@
...
+class JmapSendMessagesScenario extends Simulation {
+
+  val userCount = 100
+  val feeder = UserFeeder.createUserFeederWithInboxAndOutbox(userCount)
+  val loopVariableName = "any"
+
+  val scn = scenario("JmapSendMessages")
+    .feed(feeder)
+    .pause(1 second, 10 second)
+    .exec(authentication())
```
mbaechler: I really need to compose your scenario to avoid repetition

src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
chibenwa: Ok

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMessages.scala
chibenwa, line 37: I don't see how I can combine two jmap checks, so I wonder if it add any benefit (at least for the status, that is easily readable....)

src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
chibenwa, line 37: Nope. You've got a mix between future and List.

The closest I found was : 

```
private def createRegisteredUserFutures(userCount: Int): List[Future[User]] =
    Await.result(
      JamesWebAdministration
        .addDomain(Domain.random)
        .map(generateUsers(userCount, _)
          .map(registerUser(_))),
      Inf)
```

Not as fancy...

I think I prefer the not oneliner way to do it (even if I am a hudge one line fan !)

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
chibenwa: Ok to extract but there will be noise due to gatling typing

src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
chibenwa: Ok. Cool. Thanks.

src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
chibenwa: Yep

src/main/scala-2.11/org/apache/james/gatling/control/JamesWebAdministration.scala
chibenwa: Extracted

src/main/scala-2.11/org/apache/james/gatling/control/JamesWebAdministration.scala
chibenwa: Ok

src/test/scala-2.11/org/apache/james/gatling/jmap/scenari/JmapSendMessagesScenario.scala
chibenwa: Ok

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMessages.scala
chibenwa: Ok

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMessages.scala
chibenwa: Ok

src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
chibenwa: Ok

src/test/scala-2.11/org/apache/james/gatling/jmap/scenari/JmapSendMessagesScenario.scala
chibenwa: Ok

src/main/scala-2.11/org/apache/james/gatling/control/JamesWebAdministration.scala
```diff
@@ -2,27 +2,65 @@ package org.apache.james.gatling.control
...
     .put("")
+    .map(response => domain)
 
-  def addUser(user: User): Future[WSResponse] =
+  def addUser(user: User): Future[User] =
     wsClient.url(s"$baseUrl/users/${user.username.value}")
       .put(s"""{"password":"${user.password.value}"}""")
+      .map(response => user)
+
+  def getMailboxesUrl(username: Username): String = {
```
mbaechler: there's no URL type for that ?

src/main/scala-2.11/org/apache/james/gatling/control/JamesWebAdministration.scala
```diff
@@ -2,27 +2,65 @@ package org.apache.james.gatling.control
...
+    s"$baseUrl/users/${username.value}/mailboxes"
```
mbaechler: could it be an URL ?

src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
```diff
@@ -1,30 +1,59 @@
...
 object UserFeeder {
 
-  def createUserFeeder(userCount: Int): Array[Map[String, String]] = {
-    val domain = Domain(UUID.randomUUID().toString)
+  type UserFeeder = Array[Map[String, String]]
+
+  val USERNAME_SESSION_PARAM: String = "username"
+  val PASSWORD_SESSION_PARAM: String = "password"
+
+  def createUserFeeder(userCount: Int): UserFeeder =
```
mbaechler, line 17: why not a Future[UserFeeder] ? We could let the caller wait if it wants to, don't you think ?

src/main/scala-2.11/org/apache/james/gatling/control/JamesWebAdministration.scala
chibenwa: Yesit could be a java URL

src/test/scala-2.11/org/apache/james/gatling/jmap/scenari/common/CommonSteps.scala
```diff
@@ -0,0 +1,24 @@
...
+object CommonSteps {
+
+  def authentication(userCount: Int) =
+    scenario("JmapAuthentication")
+    .feed(UserFeeder.createUserFeederWithInboxAndOutbox(userCount))
+    .pause(1 second, 30 second)
+    .exec(JmapAuthentication.authentication())
+    .pause(1 second)
+
+  def provisionSystemMailboxes(userCount: Int) =
```
mbaechler, line 18: I think we can definitely say we will always want users provisioned with needed mailbox (and drop the other method)

src/test/scala-2.11/org/apache/james/gatling/jmap/scenari/JmapSendMessagesScenario.scala
```diff
@@ -0,0 +1,25 @@
...
+class JmapSendMessagesScenario extends Simulation {
+
+  val userCount = 100
+  val feeder = UserFeeder.createUserFeederWithInboxAndOutbox(userCount)
+  val loopVariableName = "any"
+
+  val scn = scenario("JmapSendMessages")
+    .exec(CommonSteps.provisionSystemMailboxes(userCount))
+    .repeat(360, loopVariableName) {
+      exec(JmapMessages.sendMessagesRandomly(feeder))
```
mbaechler, line 19: we don't really need a feeder, don't we ?

rouazana: it is used to take a recipient at random

src/test/scala-2.11/org/apache/james/gatling/jmap/scenari/common/CommonSteps.scala
rouazana: this method is used to retrieve the ids the the already created system mailboxes. These mailboxes are in fact created in createUserFeederWithInboxAndOutbox

src/test/scala-2.11/org/apache/james/gatling/jmap/scenari/JmapSendMessagesScenario.scala
mbaechler: of course, but you don't need a _feeder_ , an array of user is enough

src/main/scala-2.11/org/apache/james/gatling/control/UserFeeder.scala
rouazana: done, but also defined a function to not have to handle the future

src/test/scala-2.11/org/apache/james/gatling/jmap/scenari/JmapSendMessagesScenario.scala
rouazana: done

rouazana:
replaced by #17

Issue #7, opened 2016-06-09
A user should be able to read its mail


chibenwa:
A user issue GetMailboxes request.

Then a GetMessageList request.

Then a GetMessages request.

He is able to : 
- Identify the inbox from the mailboxes (and use it)
- Pick one messageId at random to later read it

PR #15, opened 2016-06-29
Issue 7 A user should be able to read its mails

chibenwa:
As a bonus, I offer flags updates.

Based https://github.com/linagora/james-gatling

Solves https://github.com/linagora/james-gatling/issues/7

### DIFF
--- a/src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMessages.scala
+++ b/src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMessages.scala
@@ -92,4 +92,26 @@ object JmapMessages {
       .check(status.is(200))
       .check(jsonPath("$.error").notExists)
 
+  def markAsRead() = performUpdate("markAsRead", "isUnread", "false")
+  def markAsAnswered() = performUpdate("markAsAnswered", "isAnswered", "true")
+  def markAsFlagged() = performUpdate("markAsFlagged", "isFlagged", "true")
+
+  def performUpdate(title: String, property: String, value: String) = {
+    JmapAuthentication.authenticatedQuery(title, "/jmap")
+      .body(StringBody(
+        s"""[[
+          "setMessages",
+          {
+            "update": {
+              "$${messageIds.random()}" : {
+                "$property": "$value"
+              }
+            }
+          },
+          "#0"
+          ]]"""))
+      .check(status.is(200))
+      .check(jsonPath("$.error").notExists)
+  }
+
 }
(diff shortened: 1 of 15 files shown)


src/test/scala-2.11/org/apache/james/gatling/jmap/scenari/JmapGetMessageListScenario.scala
```diff
@@ -0,0 +1,36 @@
...
+import scala.concurrent.duration._
+
+class JmapGetMessageListScenario extends Simulation {
+
+
+  val userCount = 100
+
+  val feeder = UserFeeder.createUserFeederWithInboxAndOutbox(userCount)
+
+  val scn = scenario("JmapGetMessagesLists")
```
mbaechler, line 17: scenario are composable, please do not copy/paste from previous scenario

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMessages.scala
```diff
@@ -60,4 +60,35 @@ object JmapMessages {
       .check(jsonPath("$.error").notExists)
       .check(jsonPath("$[0][1].messageIds[*]").findAll.saveAs("messageIds"))
 
+  def getRandomMessage() =
+    JmapAuthentication.authenticatedQuery("getMessages", "/jmap")
+      .body(StringBody(
+        """[[
+          "getMessages",
+          {
+            "ids": ["${messageIds.random()}"],
```
mbaechler, line 71: please, no code in strings

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMessages.scala
```diff
@@ -91,4 +91,26 @@ object JmapMessages {
       .check(status.is(200))
       .check(jsonPath("$.error").notExists)
 
+  def markAsRead() = performUpdate("markAsRead", "isUnread", "false")
```
mbaechler: strings everywhere ! you can do much better, use types, compose, etc

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMessages.scala
chibenwa, line 71: It is not scala code. It is not a s"" string.

Here you see Gatling DSL on session access. 

By the way, there is no other way to do it.

rouazana:
rebased here: #20

PR #17, opened 2016-09-21
Issue 6 I should be able to send a mail randomly using gatling

rouazana:
Follow #14

### DIFF
--- a/src/main/scala-2.11/org/apache/james/gatling/control/UserCreator.scala
+++ b/src/main/scala-2.11/org/apache/james/gatling/control/UserCreator.scala
@@ -1,0 +1,38 @@
+package org.apache.james.gatling.control
+
+import io.gatling.core.Predef._
+
+import scala.concurrent.duration.Duration.Inf
+import scala.concurrent.{Await, Future}
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.util.Success
+import scala.concurrent.Awaitable
+
+object UserCreator {
+
+  def createUsersWithInboxAndOutbox(userCount: Int): Seq[Future[User]] =
+        createUsers(userCount)
+          .map(userFuture => userFuture.andThen {
+            case Success(user) => registerSystemMailboxes(user)
+          })
+
+  private def createUsers(userCount: Int): Seq[Future[User]] = {
+    val domain = Domain.random
+    JamesWebAdministration.addDomain(domain).get
+
+    generateUsers(userCount, domain)
+      .map(JamesWebAdministration.addUser)
+  }
+
+  private def generateUsers(userCount: Int, domain: Domain): Seq[User] =
+    (0 until userCount)
+      .map(i => User.random(domain))
+
+  def registerSystemMailboxes(user: User): Future[User] = {
+    Future.sequence(
+      List(JamesWebAdministration.createInbox(user.username),
+        JamesWebAdministration.createOutbox(user.username),
+        JamesWebAdministration.createSentBox(user.username)))
+      .map(responseList => user)
+  }
+}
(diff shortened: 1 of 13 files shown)


src/test/scala-2.11/org/apache/james/gatling/jmap/scenari/JmapSendMessagesScenario.scala
```diff
@@ -6,17 +6,19 @@ import org.apache.james.gatling.jmap.scenari.common.CommonSteps
 import org.apache.james.gatling.jmap.{HttpSettings, JmapMessages}
 
 import scala.concurrent.duration._
+import org.apache.james.gatling.control.{Username, Password, User}
 
 class JmapSendMessagesScenario extends Simulation {
 
   val userCount = 100
-  val feeder = UserFeeder.createUserFeederWithInboxAndOutbox(userCount)
+  val users = UserFeeder.createCompletedUserFeederWithInboxAndOutbox(userCount)
```
mbaechler: what if we create a function that return the array for users and we map it to a feeder when we need it ?
It sounds more logical, don't you think ?

rouazana: I tried something, WDYT?
At least the code seems cleaner.

src/main/scala-2.11/org/apache/james/gatling/control/UserCreator.scala
```diff
@@ -0,0 +1,39 @@
...
+    JamesWebAdministration.addDomain(domain).get
+
+    generateUsers(userCount, domain)
+      .map(JamesWebAdministration.addUser)
+  }
+
+  private def generateUsers(userCount: Int, domain: Domain): List[User] =
+    (0 until userCount)
+      .map(i => User.random(domain))
+      .toList
```
mbaechler: why do you need a list ?

src/main/scala-2.11/org/apache/james/gatling/control/UserCreator.scala
```diff
@@ -0,0 +1,39 @@
...
+
+  def createUsersWithInboxAndOutbox(userCount: Int): List[Future[User]] =
+        createUsers(userCount)
+          .map(userFuture => userFuture.andThen {
+            case Success(user) => registerSystemMailboxes(user)
+          })
+
+  private def createUsers(userCount: Int): List[Future[User]] = {
+    val domain = Domain.random
+    JamesWebAdministration.addDomain(domain).get
```
mbaechler, line 21: what about doing a map here ?

src/test/scala-2.11/org/apache/james/gatling/jmap/scenari/JmapSendMessagesScenario.scala
```diff
@@ -0,0 +1,27 @@
...
+import org.apache.james.gatling.jmap.scenari.common.CommonSteps
+import org.apache.james.gatling.jmap.{HttpSettings, JmapMessages}
+
+import scala.concurrent.duration._
+import org.apache.james.gatling.control.{Username, Password, User}
+
+class JmapSendMessagesScenario extends Simulation {
+
+  val userCount = 100
+  val users = UserFeeder.createCompletedUserFeederWithInboxAndOutbox(userCount)
```
mbaechler: you no longer need the feeder here, don't you ?

rouazana: yes, I was working on it

src/main/scala-2.11/org/apache/james/gatling/control/UserCreator.scala
rouazana: replaced (everywhere) by a Seq

src/main/scala-2.11/org/apache/james/gatling/control/UserCreator.scala
rouazana: and returning Future[List[Future[User]]? That seems to much future for me :)

mbaechler (approved):

chibenwa (approved):

PR #27, opened 2016-10-11
Issue 11 reauthentication POC

rouazana:
Here is a reauthentication POC, used on the simpler method: GetMailboxes.
Please check if it's enough fine to be generalized.

### DIFF
--- a/src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
+++ b/src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
@@ -22,6 +22,8 @@ object JmapMailboxes {
 
   def getSystemMailboxes = getMailboxes
 
+  def getSystemMailboxesWithChecks = getSystemMailboxes.check(getSystemMailboxesChecks: _*)
+
   val getSystemMailboxesChecks: Seq[HttpCheck] = getMailboxesChecks ++ List[HttpCheck](
     jsonPath(inboxIdPath).saveAs("inboxMailboxId"),
     jsonPath(outboxIdPath).saveAs("outboxMailboxId"),
--- a/src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
+++ b/src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
@@ -3,30 +3,27 @@ package org.apache.james.gatling.jmap
 import io.gatling.core.Predef._
 import io.gatling.http.Predef._
 import org.apache.james.gatling.utils.JmapChecks
+import io.gatling.http.check.HttpCheck
 
 object JmapMailboxes {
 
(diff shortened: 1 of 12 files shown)


src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
```diff
@@ -3,30 +3,27 @@ package org.apache.james.gatling.jmap
...
-  def getSystemMailboxes = {
-    val inboxIdPath = s"$$$mailboxListPath[?(@.role == 'inbox')].id"
-    val outboxIdPath = s"$$$mailboxListPath[?(@.role == 'outbox')].id"
-    val sentIdPath = s"$$$mailboxListPath[?(@.role == 'sent')].id"
-    getMailboxes
-      .check(jsonPath(inboxIdPath).saveAs("inboxMailboxId"))
-      .check(jsonPath(outboxIdPath).saveAs("outboxMailboxId"))
-      .check(jsonPath(sentIdPath).saveAs("sentMailboxId"))
-  }
+  def getSystemMailboxes = getMailboxes
```
chibenwa, line 23: Does line 23 imples all mailbox is system mailbox ?

chibenwa (commented):
Overall, I agree with this POC
Just, in my opinion, to be a complete :+1:, I would remove now useless getSystemMailbox method....

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
rouazana: Of course not. It just that it uses the same request.
I wanted to keep one request / one checks set for each available request, but I can remove it if you find it better

chibenwa (approved):

mbaechler (approved):

aduprat (approved):

src/main/scala-2.11/org/apache/james/gatling/utils/RetryAuthentication.scala
```diff
@@ -0,0 +1,24 @@
...
+
+  def execWithRetryAuthentication(scenario: HttpRequestBuilder, checks: Seq[HttpCheck]): ChainBuilder =
+    exec(scenario.check(status.in(200, 401).saveAs("statusCode")).check(checkIfOk(checks): _*))
+      .doIfEquals("${statusCode}", 401){
+          JmapAuthentication.authentication()
+            .exec(scenario.check(checks: _*))
+      }
+
+  def checkIfOk(checks: Seq[HttpCheck]): Seq[HttpCheck] =
+    checks.map(check => checkIf((r: Response, s: Session) => r.statusCode == 200){check})
```
chibenwa, line 22: r.statusCode is an Option, and not an int. Hence, it is always false and test were never performed....

chibenwa:
Move here : https://github.com/linagora/james-gatling/pull/29

Issue #34, opened 2017-06-14
SmtpAction should rely on the action system


chibenwa:
Right now it's not the case. The mail is directly sent in the action. Thus we quickly encounter a bottleneck when playing load tests with Gatling SMTP.

Note that reports are also perturbated from this, and not issued every 5 seconds.

Note that it makes no difference using 100 users and 1000 users.

What we should be doing is relying on the actor system for sending emails. We can also plan using some asynchronous SMTP librairies.

chibenwa:
Solved

PR #35, opened 2017-06-14
ISSUE-34 Gatling SMTP should be asynchronous and rely on the event system

chibenwa:
Solves https://github.com/linagora/james-gatling/issues/34

One minute with 10 users, sending 1 mail per second:

10_users

One minute with 100 users, sending 1 mail per second:

100_users

We clearly see that now Gatling can quickly overwhelm James, what was not possible before due to synchronous behaviour.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,7 +8,10 @@ enablePlugins(GatlingPlugin)
 
 EclipseKeys.withSource := true
 
+resolvers += "lightshed-maven" at "http://dl.bintray.com/content/lightshed/maven"
+
 libraryDependencies += "com.typesafe.play" %% "play-ws" % "2.4.3"
 libraryDependencies += "io.gatling" % "gatling-test-framework" % "2.2.2"
 libraryDependencies += "io.gatling.highcharts" % "gatling-charts-highcharts" % "2.2.2"
-libraryDependencies += "org.apache.commons" % "commons-email" % "1.3.2"
+
+libraryDependencies += "ch.lightshed" %% "courier" % "0.1.4"
--- a/src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpAction.scala
+++ b/src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpAction.scala
@@ -23,8 +23,12 @@ class SmtpAction(requestName: String,
   val name = "sendMail"
 
   def execute(session: Session) {
-    def readSession(variableName: String): String = session.apply(variableName).as[String]
-    val request = SendMailRequest(session = session,
+    context.actorOf(SmtpHandler.props()) ! generateSendMailRequest(session)
+  }
+
+  private def generateSendMailRequest(session: Session) = {
+    def readSession(variableName: String) = session(variableName).as[String]
+    SendMailRequest(session = session,
       host = protocol.host,
       port = protocol.port,
       ssl = protocol.ssl,
@@ -32,22 +36,18 @@ class SmtpAction(requestName: String,
       to = readSession(UserFeeder.UsernameSessionParam),
       subject = subject,
       body = body,
-      credentials = readCredentials(readSession)(protocol))
-
-    context.actorOf(SmtpHandler.props()) ! request
-
-    next ! session
+      credentials = provideCredentialsIfNeeded(readSession)(protocol))
   }
 
   override def receive: Receive = {
     case session: Session => execute(session)
     case executionReport: ExecutionReport =>
       statsEngine.logResponse(executionReport.session, requestName, executionReport.responseTimings, executionReport.status, None, executionReport.errorMessage)
-    case Failure(e) =>
-      logger.error("Exception caught while sending mail", e)
+      next ! executionReport.session
+    case msg => logger.error(s"Unexpected message $msg")
   }
 
-  def readCredentials(sessionReader: String => String)(protocol: SmtpProtocol): Option[Credentials] = {
+  def provideCredentialsIfNeeded(sessionReader: String => String)(protocol: SmtpProtocol): Option[Credentials] = {
     if (protocol.auth) None
     else Some(Credentials(sessionReader(UserFeeder.UsernameSessionParam), sessionReader(UserFeeder.PasswordSessionParam)))
   }
--- a/src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpAction.scala
+++ b/src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpAction.scala
@@ -6,8 +6,6 @@ import io.gatling.core.session.Session
 import io.gatling.core.stats.StatsEngine
 import org.apache.james.gatling.control.UserFeeder
 
-import scala.util.Failure
-
 object SmtpAction {
   def props(requestName: String, subject: String, body: String, statsEngine: StatsEngine, next: Action, protocol: SmtpProtocol) =
     Props(new SmtpAction(requestName, subject, body, statsEngine, next, protocol))
@@ -20,10 +18,12 @@ class SmtpAction(requestName: String,
                   val next: Action,
                   protocol: SmtpProtocol) extends ChainableAction with Actor {
 
+  val smtpHandler = context.actorOf(SmtpHandler.props())
+
   val name = "sendMail"
 
   def execute(session: Session) {
-    context.actorOf(SmtpHandler.props()) ! generateSendMailRequest(session)
+    smtpHandler ! generateSendMailRequest(session)
   }
 
   private def generateSendMailRequest(session: Session) = {
--- a/src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpAction.scala
+++ b/src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpAction.scala
@@ -1,70 +1,54 @@
 package org.apache.james.gatling.smtp
 
-import io.gatling.commons.stats.{KO, OK}
-import io.gatling.core.Predef._
+import akka.actor.{Actor, Props}
 import io.gatling.core.action.{Action, ChainableAction}
 import io.gatling.core.session.Session
 import io.gatling.core.stats.StatsEngine
-import io.gatling.core.stats.message.ResponseTimings
-import org.apache.commons.mail.SimpleEmail
 import org.apache.james.gatling.control.UserFeeder
 
-import scala.util.{Failure, Success, Try}
+import scala.util.Failure
 
-class SmtpAction(
-                  requestName: String,
+object SmtpAction {
+  def props(requestName: String, subject: String, body: String, statsEngine: StatsEngine, next: Action, protocol: SmtpProtocol) =
+    Props(new SmtpAction(requestName, subject, body, statsEngine, next, protocol))
+}
+
+class SmtpAction(requestName: String,
                   subject: String,
                   body: String,
                   val statsEngine: StatsEngine,
                   val next: Action,
-                  protocol: SmtpProtocol) extends ChainableAction {
+                  protocol: SmtpProtocol) extends ChainableAction with Actor {
 
   val name = "sendMail"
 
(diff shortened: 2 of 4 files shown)


src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpAction.scala
```diff
@@ -1,70 +1,54 @@
...
   val name = "sendMail"
 
   def execute(session: Session) {
-    val email: SimpleEmail = generateEmail(session)
-    val reqStart = System.currentTimeMillis()
-    val executionReport = sendMail(email)
-    val reqEnd = System.currentTimeMillis()
-
-    statsEngine.logResponse(session, requestName, ResponseTimings(reqStart, reqEnd), executionReport.status, None, executionReport.errorMessage)
+    def readSession(variableName: String): String = session.apply(variableName).as[String]
```
mbaechler: `def readSession(variableName: String) = session(variableName).as[String]` is enough IMO

src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpAction.scala
```diff
@@ -1,70 +1,54 @@
...
+    val request = SendMailRequest(session = session,
```
mbaechler: generateEmail was a good name for that, I would not inline that code

src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpAction.scala
```diff
@@ -1,70 +1,54 @@
...
-    email.setMsg(body)
-    email.setFrom(userEmail)
-    email.addTo(userEmail)
-    email
+  override def receive: Receive = {
+    case session: Session => execute(session)
+    case executionReport: ExecutionReport =>
+      statsEngine.logResponse(executionReport.session, requestName, executionReport.responseTimings, executionReport.status, None, executionReport.errorMessage)
+    case Failure(e) =>
+      logger.error("Exception caught while sending mail", e)
```
mbaechler: there's nothing to report in that case ?

src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpAction.scala
```diff
@@ -1,70 +1,54 @@
...
 
-  private def sendMail(email: SimpleEmail): ExecutionReport = {
-    Try(email.send()) match {
-      case Success(v) => GoodExecutionReport()
-      case Failure(e) =>
-        logger.error("Exception caught while sending mail", e)
-        BadExecutionReport(e.getMessage)
-    }
+  def readCredentials(sessionReader: String => String)(protocol: SmtpProtocol): Option[Credentials] = {
+    if (protocol.auth) None
```
mbaechler: it's not what I expect readCredentials to do, maybe you could change method name to reflect that logic ?

src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpHandler.scala
```diff
@@ -0,0 +1,80 @@
...
+      context.stop(self)
+    case msg =>
+      logger.error(s"received unexpected message $msg")
+  }
+
+  def send(sendMailRequest: SendMailRequest): ExecutionReport = doSend(generateEmail(sendMailRequest), sendMailRequest.session)
+
+  private def generateEmail(sendMailRequest: SendMailRequest): SimpleEmail = {
+    val email = new SimpleEmail()
+    sendMailRequest.credentials.fold {} {value => email.setAuthentication(value.login, value.password)}
```
mbaechler: this line is a bit cryptic, what does the empty block after fold ? (there's many version of fold ...)

src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpHandler.scala
```diff
@@ -0,0 +1,80 @@
...
+    email.setHostName(sendMailRequest.host)
+    email.setSSLOnConnect(sendMailRequest.ssl)
+    email.setSmtpPort(sendMailRequest.port)
+    email.setSubject(sendMailRequest.subject + " [" + Math.random() + "]")
```
mbaechler: why this random is not in the request?

build.sbt
```diff
@@ -8,7 +8,10 @@ enablePlugins(GatlingPlugin)
...
 EclipseKeys.withSource := true
 
+resolvers += "lightshed-maven" at "http://dl.bintray.com/content/lightshed/maven"
+
 libraryDependencies += "com.typesafe.play" %% "play-ws" % "2.4.3"
 libraryDependencies += "io.gatling" % "gatling-test-framework" % "2.2.2"
 libraryDependencies += "io.gatling.highcharts" % "gatling-charts-highcharts" % "2.2.2"
-libraryDependencies += "org.apache.commons" % "commons-email" % "1.3.2"
+
+libraryDependencies += "ch.lightshed" %% "courier" % "0.1.4"
```
mbaechler, line 17: very great finding ! we definitively need to contribute to that project

src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpHandler.scala
```diff
@@ -17,39 +18,49 @@ object SmtpHandler {
...
+
     val requestStart = System.currentTimeMillis()
-    Try(email.send()) match {
-      case Success(v) => GoodExecutionReport(computeResponseTimings(requestStart), session)
-      case Failure(e) =>
-        logger.error("Exception caught while sending mail", e)
-        BadExecutionReport(e.getMessage, computeResponseTimings(requestStart), session)
-    }
+
+    val future = mailer.apply(Envelope.from(new InternetAddress(sendMailRequest.from))
```
mbaechler: `.apply` is not meant to be call, you can just use `()`

src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpHandler.scala
```diff
@@ -17,39 +18,49 @@ object SmtpHandler {
...
+      .to(new InternetAddress(sendMailRequest.to))
+      .subject(sendMailRequest.subject)
+      .content(Text(sendMailRequest.body)))(ExecutionContext.global)
```
mbaechler: you are supposed to import it to benefits from implicit arg

src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpHandler.scala
```diff
@@ -1,14 +1,15 @@
...
 import io.gatling.core.Predef.Status
 import io.gatling.core.akka.BaseActor
 import io.gatling.core.session.Session
 import io.gatling.core.stats.message.ResponseTimings
-import org.apache.commons.mail.SimpleEmail
 
-import scala.util.{Failure, Success, Try}
+import scala.concurrent.ExecutionContext
 
 object SmtpHandler {
```
mbaechler: do you instantiate an actor for each email or you only use one in the system ?

chibenwa: For each email

src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpHandler.scala
```diff
@@ -17,39 +18,49 @@ object SmtpHandler {
...
+    future.onFailure {case e =>
+      logger.error("Exception caught while sending mail", e)
+      self ! BadExecutionReport(e.getMessage, computeResponseTimings(requestStart), sendMailRequest.session)
+    }(ExecutionContext.global)
   }
 
   private def computeResponseTimings(reqStart: Long) = {
     val requestEnd = System.currentTimeMillis()
+    val duration = requestEnd - reqStart
+    println(s"$reqStart -> $requestEnd = $duration")
```
rouazana: debug statement?

chibenwa: Yes sorry.

src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpAction.scala
chibenwa: What do you mean by report? To who?

If the execution failed for any reason, the failure is handled by a failed execution report. Here the error is for our code.

src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpHandler.scala
chibenwa: I will remove...

src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpHandler.scala
chibenwa: Cool! Thanks

src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpAction.scala
rouazana: you don't need to log the failiure in statsEngine?

chibenwa: Even if it's a failure on gatling side ?

chibenwa: I mean there is no context, maybe this switch is not needed as error is handled in the handler.

rouazana: I think I see your point. The (normal) error should have been handled by the previous case, as an ExecutionReport can be a Failure or a Success. So if we get a Failure here, logging should be sufficient as it should be an internal Gatling error.

src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpHandler.scala
mbaechler: why ? I mean, if the mailer is async, you should not need that ?

src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpAction.scala
chibenwa: I chose **provideCredentialsIfNeeded**

src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpHandler.scala
chibenwa: Won't fix: this code is outdated...

src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpHandler.scala
chibenwa: magic!

src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpAction.scala
chibenwa: I removed the failure case

src/main/scala-2.11/org/apache/james/gatling/smtp/SmtpHandler.scala
```diff
@@ -0,0 +1,89 @@
...
+                           ssl: Boolean,
+                           from: String,
+                           to: String,
+                           subject: String,
+                           body: String,
+                           credentials: Option[Credentials])
+
+case class Credentials(login: String, password: String)
+
+abstract class ExecutionReport(_errorMessage: Option[String], _status: Status, _responseTimings: ResponseTimings, _session: Session) {
```
mbaechler: what are these underscores everywhere ?!

mbaechler (approved):

aduprat (approved):

PR #36, opened 2017-12-05
Big set scenario

aduprat:
The checks of the scenario may be enhanced.

### DIFF
--- a/src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
+++ b/src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
@@ -5,6 +5,10 @@ import io.gatling.http.Predef._
 import io.gatling.http.check.HttpCheck
 import org.apache.james.gatling.utils.JmapChecks
 import org.apache.james.gatling.utils.RetryAuthentication.execWithRetryAuthentication
+import org.apache.james.gatling.utils.RandomStringGenerator
+
+case class Id(id: String = RandomStringGenerator.randomString)
+case class Name(name: String = RandomStringGenerator.randomString)
 
 object JmapMailboxes {
 
@@ -16,6 +20,17 @@ object JmapMailboxes {
   def getMailboxes =
     JmapAuthentication.authenticatedQuery("getMailboxes", "/jmap")
       .body(StringBody("""[["getMailboxes", {}, "#0"]]"""))
+
+  def createMailbox(id: Id, name: Name) =
+    JmapAuthentication.authenticatedQuery("setMailboxes", "/jmap")
+      .body(StringBody(s"""[["setMailboxes", 
+            {
+              "create": {
+                "${id.id}": {
+                  "name": "${name.name}"
+                } 
+              }
+            }, "#0"]]"""))
 
   val getMailboxesChecks: Seq[HttpCheck] = List(
     status.is(200),
--- a/src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
+++ b/src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
@@ -7,8 +7,17 @@ import org.apache.james.gatling.utils.JmapChecks
 import org.apache.james.gatling.utils.RetryAuthentication.execWithRetryAuthentication
 import org.apache.james.gatling.utils.RandomStringGenerator
 
-case class Id(id: String = RandomStringGenerator.randomString)
-case class Name(name: String = RandomStringGenerator.randomString)
+object IdFactory {
+  def apply(): Id =
+    Id(RandomStringGenerator.randomString)
+}
+case class Id private(id: String)
+
+object NameFactory {
+  def apply(): Name =
+    Name(RandomStringGenerator.randomString)
+}
+case class Name private(name: String)
 
 object JmapMailboxes {
 
--- a/src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
+++ b/src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
@@ -21,6 +21,14 @@ object JmapMailboxes {
     JmapAuthentication.authenticatedQuery("getMailboxes", "/jmap")
       .body(StringBody("""[["getMailboxes", {}, "#0"]]"""))
 
+  def getMailboxIdByName(name: Name) =
+    JmapAuthentication.authenticatedQuery("getMailboxes", "/jmap")
+      .body(StringBody("""[["getMailboxes", {}, "#0"]]"""))
+      .check(getMailboxIdCheck(name))
+
+  def getMailboxIdCheck(name: Name) =
+      jsonPath(s"$$$mailboxListPath[?(@.name == '${name.name}')].id").saveAs("mailboxId")
+      
   def createMailbox(id: Id, name: Name) =
     JmapAuthentication.authenticatedQuery("setMailboxes", "/jmap")
       .body(StringBody(s"""[["setMailboxes", 
--- a/src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
+++ b/src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
@@ -31,11 +31,10 @@ object JmapMailboxes {
       .body(StringBody("""[["getMailboxes", {}, "#0"]]"""))
 
   def getMailboxIdByName(name: Name) =
-    JmapAuthentication.authenticatedQuery("getMailboxes", "/jmap")
-      .body(StringBody("""[["getMailboxes", {}, "#0"]]"""))
-      .check(getMailboxIdCheck(name))
+    getMailboxes
+      .check(retrieveMailboxIdByName(name))
 
-  def getMailboxIdCheck(name: Name) =
+  def retrieveMailboxIdByName(name: Name) =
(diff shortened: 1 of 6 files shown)


src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
```diff
@@ -5,6 +5,10 @@ import io.gatling.http.Predef._
 import io.gatling.http.check.HttpCheck
 import org.apache.james.gatling.utils.JmapChecks
 import org.apache.james.gatling.utils.RetryAuthentication.execWithRetryAuthentication
+import org.apache.james.gatling.utils.RandomStringGenerator
+
+case class Id(id: String = RandomStringGenerator.randomString)
```
mbaechler: it should be related to a Mailbox case class IMO.

mbaechler: you should not use the default parameter for that but use a factory method instead

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
```diff
@@ -21,6 +21,14 @@ object JmapMailboxes {
     JmapAuthentication.authenticatedQuery("getMailboxes", "/jmap")
       .body(StringBody("""[["getMailboxes", {}, "#0"]]"""))
 
+  def getMailboxIdByName(name: Name) =
+    JmapAuthentication.authenticatedQuery("getMailboxes", "/jmap")
+      .body(StringBody("""[["getMailboxes", {}, "#0"]]"""))
+      .check(getMailboxIdCheck(name))
+
+  def getMailboxIdCheck(name: Name) =
```
mbaechler: it's not a getter, it extracts and saves

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
```diff
@@ -21,6 +21,14 @@ object JmapMailboxes {
     JmapAuthentication.authenticatedQuery("getMailboxes", "/jmap")
       .body(StringBody("""[["getMailboxes", {}, "#0"]]"""))
 
+  def getMailboxIdByName(name: Name) =
+    JmapAuthentication.authenticatedQuery("getMailboxes", "/jmap")
```
mbaechler: it's a copy/paste of getMailboxes

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMessages.scala
```diff
@@ -42,6 +44,24 @@ object JmapMessages {
           "#0"
           ]]"""))
   
+  def getMessageIds(mailboxId: String) =
```
mbaechler: still not a getter

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMessages.scala
```diff
@@ -62,6 +62,23 @@ object JmapMessages {
     jsonPath(messageIdsPath).findAll.saveAs("messageIds")
   }
 
+  def moveSentMessagesToMailboxId = 
+    foreach("${messageIds}", "messageId") {
```
mbaechler: how do you know messageIds attribute is sent messages ?

mbaechler: why do you want to issue a query for each message ?

src/test/scala-2.11/org/apache/james/gatling/jmap/scenari/common/CommonSteps.scala
```diff
@@ -36,6 +39,23 @@ object CommonSteps {
       }
       .pause(30 second)
 
+  def provisionUsersWithMailboxesAndMessages(users: Seq[Future[User]]) =
+    scenario("ProvisionUsersWithMailboxesAndMessages")
+      .exec(provisionSystemMailboxes(users))
+      .repeat(NumberOfMailboxes) {
```
mbaechler: please pass it in parameter

src/test/scala-2.11/org/apache/james/gatling/jmap/scenari/JmapBigSetScenario.scala
```diff
@@ -0,0 +1,34 @@
...
+
+/*
+ * The aim of the scenario is to provide multiple mailboxes per user and several mails in them.
+ * Then, it's just checking that the created mailboxes exists.
+ */
+class JmapBigSetScenario extends Simulation {
+
+  val users = new UserCreator(BaseJamesWebAdministrationUrl).createUsersWithInboxAndOutbox(UserCount)
+  
+  val scn = scenario("JmapBigSet")
```
mbaechler, line 28: you don't check mailboxes ?

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
strokyl: I don't understand this function on what json does jsonPath work on ?

quynhn (commented):
:+1:

src/test/scala-2.11/org/apache/james/gatling/jmap/scenari/common/CommonSteps.scala
```diff
@@ -36,6 +39,23 @@ object CommonSteps {
...
+        var mailboxName = Name()
+        exec(JmapMailboxes.createMailbox(Id(), mailboxName))
+          .pause(1 second, 2 seconds)
```
strokyl: What does that mean you pause for 1 second, 2 seconds or 3 seconds. Or something random between 1 and 2 seconds ?

chibenwa: Random between 1 and 2

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
aduprat: There's the same pattern in JmapMessages, I think the whole project needs some refactoring.

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMessages.scala
aduprat: That's not really easy, _foreach_ is a gatling method & I don't succeed to chain to something else than gatling methods.

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
```diff
@@ -5,6 +5,19 @@ import io.gatling.http.Predef._
 import io.gatling.http.check.HttpCheck
 import org.apache.james.gatling.utils.JmapChecks
 import org.apache.james.gatling.utils.RetryAuthentication.execWithRetryAuthentication
+import org.apache.james.gatling.utils.RandomStringGenerator
+
+object IdFactory {
```
mbaechler: I would call it `Id` and call the method `generate`

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMessages.scala
```diff
@@ -42,6 +44,41 @@ object JmapMessages {
           "#0"
           ]]"""))
   
+  def retrieveMessageIds(mailboxId: String) =
```
mbaechler: don't we have a Mailbox.Id type ?

strokyl (approved):

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMessages.scala
```diff
@@ -42,6 +44,41 @@ object JmapMessages {
...
+          "#0"
+          ]]"""))
+      .check(saveMessageIds)    
+
+  def saveMessageIds = {
+    jsonPath(messageIdsPath).findAll.saveAs("messageIds")
+  }
+
+  def moveMessagesToMailboxId =
+    foreach("${messageIds}", "messageId") {
```
mbaechler: something like that (did not compile, it's just pseudocode) : 
```
exec((s: Session) => s.setAttribute("update", 
  s("messageIds").map(x => """
  "$${x}" : {
                  "mailboxIds": [ "$${mailboxId}" ]"""
).join(","))
.exec(JmapAuthentication.authenticatedQuery("moveSentMessagesToMessageId", "/jmap")
        .body(StringBody(
          s"""[[
            "setMessages",
            {
              "update": {
                "$${update}
                }
              }
            },
            "#0"
            ]]""")))
```

aduprat: Thx

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMessages.scala
```diff
@@ -63,21 +64,22 @@ object JmapMessages {
...
-                "$${messageId}" : {
-                  "mailboxIds": [ "$${mailboxId}" ]
-                }
-              }
-            },
-            "#0"
-            ]]""")))
-    }
+    exec((session: Session) => session.set("update", {
+        var mailboxId = session("mailboxId").as[String]
```
mbaechler: val !!!!

aduprat:
#37

PR #38, opened 2017-12-18
Introduce a scenario checking the number of automatically provisioned…

aduprat:
… system mailboxes


chibenwa:
That's cool but.... Ghost mailbox bug will not remove mailboxes. You will only notice changing IDs. 

What you could be doing is saving the id of each mailbox inside the Gatling session, and compare the returned value with the session one.

quynhn:
Could we close this one and review all on https://github.com/linagora/james-gatling/pull/39?

PR #39, opened 2017-12-28
MailboxId of system mailboxes should not be changed scenario

quynhn:


chibenwa:
Could we clean the history? I don't expect we keep a WIP commit...

src/main/scala-2.11/org/apache/james/gatling/jmap/JmapMailboxes.scala
```diff
@@ -60,10 +62,23 @@ object JmapMailboxes {
   def getMailboxesChecks(expectedNumberOfMailboxes: Int): Seq[HttpCheck] =
     getMailboxesChecks :+ assertNumberOfMailboxes(expectedNumberOfMailboxes)
 
+  def checkMailboxIdsHasNotChange: Seq[HttpCheck] =
```
chibenwa: We should indicate it is system mailboxes, no?

Like `checkSystemMailboxIdsHaveNotChange`

aduprat (approved):

PR #40, opened 2018-07-11
GITLAB-488 Check the state of the response before saving message ids 

aduprat:


src/test/scala-2.11/org/apache/james/gatling/jmap/scenari/JmapCountMailboxesScenario.scala
```diff
@@ -25,8 +25,5 @@ class JmapCountMailboxesScenario extends Simulation {
         .pause(1 second , 2 seconds)
     }
 
-  setUp(
-    scn.inject(
-      rampUsers(1000) over(5 minutes)
-  )).protocols(HttpSettings.httpProtocol)
+  setUp(scn.inject(atOnceUsers(UserCount))).protocols(HttpSettings.httpProtocol)
```
rouazana, line 28: why did you remove the ramping?

aduprat: 1/ To have the same injection as other scenari
2/ It's failing if you play this scenario with a configuration of less than 1000 users

rouazana (approved):

chibenwa (approved):

PR #43, opened 2018-11-26
 Introduce a basic smtp test around a csv file feeder 

aduprat:
The first part of this PR is to trust all ssl certificates.

The second part of this PR is to enable smtp stress test on a ldap product base of James.
(this means that we are not able to provision users)


src/test/scala-2.11/org/apache/james/gatling/smtp/scenari/FeederNoAuthenticationNoEncryptionScenario.scala
```diff
@@ -0,0 +1,25 @@
...
+import org.apache.james.gatling.jmap.scenari.common.Configuration._
+import org.apache.james.gatling.smtp.scenari.common.Configuration._
+
+import scala.concurrent.duration._
+
+class FeederNoAuthenticationNoEncryptionScenario extends Simulation {
+
+  def feeder = csv("users.csv")
+
+  val scn = scenario("SMTP_No_Authentication_No_Encryption")
```
blackheaven: Maybe the name has to be changed

trantienduchn (approved):

src/test/scala-2.11/org/apache/james/gatling/smtp/scenari/common/Configuration.scala
```diff
@@ -0,0 +1,10 @@
+package org.apache.james.gatling.smtp.scenari.common
+
+import org.apache.james.gatling.smtp.{SmtpActionBuilder, SmtpProtocolBuilder}
+
+object Configuration {
+
+  val smtp = SmtpProtocolBuilder.default.build()
+
+  def smtp(requestName: String) = new SmtpActionBuilder(requestName, null, null)
```
mbaechler, line 9: I don't think it's related to a scenario, it's part of smtp protocol implementation, isn't it ?

src/test/scala-2.11/org/apache/james/gatling/smtp/scenari/FeederNoAuthenticationNoEncryptionScenario.scala
```diff
@@ -0,0 +1,25 @@
...
+    .feed(feeder)
```
mbaechler, line 15: why do you need to copy/paste the scenario just to change the feeder ?

src/test/scala-2.11/org/apache/james/gatling/smtp/scenari/common/Configuration.scala
aduprat: Because the SMTP server you are willing to test may not have a standard configuration (ssl, other port, ...), so it really depend on the execution of the test.
It's seems legit to have such configuration in the `test` package.

src/test/scala-2.11/org/apache/james/gatling/smtp/scenari/FeederNoAuthenticationNoEncryptionScenario.scala
aduprat: 1/ I think it's cool to keep the main steps in the scenario; that way you are able to understand what part of the server will be stressed while playing it.

2/ It's only 10 lines of code.

mbaechler (approved):

blackheaven (approved):

PR #47, opened 2018-12-19
Add JmapGetMailboxesHugeNumberScenario

blackheaven:


rouazana (approved):
that's not so "huge" but :+1:

blackheaven:
That's always 67 times larger :-°

src/main/scala-2.11/org/apache/james/gatling/control/JamesWebAdministration.scala
```diff
@@ -45,18 +45,20 @@ class JamesWebAdministration(val baseUrl: URL) {
...
   }
 
   def createSentBox(username: Username) = {
+    createMailbox(username, "Sent")
+  }
+
+  def createMailbox(username: Username, name: String) = {
     val mailboxesUrl = getMailboxesUrl(username)
-    wsClient.url(s"$mailboxesUrl/Sent").put("")
+    wsClient.url(s"$mailboxesUrl/" + name).put("")
```
mbaechler: come on, that's scala, you can write `s"$mailboxesUrl/$name`

src/main/scala-2.11/org/apache/james/gatling/control/UserCreator.scala
```diff
@@ -35,4 +41,10 @@ class UserCreator(val baseJamesWebAdministrationUrl: URL) {
         jamesWebAdministration.createOutbox(user.username),
         jamesWebAdministration.createSentBox(user.username)))
       .map(responseList => user)
+
+  def registerOtherMailboxes(user: User, mailboxCount: Int): Future[User] =
+    Future.sequence(
+      (1.to(mailboxCount)
+        .map(index => jamesWebAdministration.createMailbox(user.username, "MBOX" + index.toString))))
```
mbaechler: what about using string interpolation ?

src/main/scala-2.11/org/apache/james/gatling/control/JamesWebAdministration.scala
```diff
@@ -45,18 +45,20 @@ class JamesWebAdministration(val baseUrl: URL) {
...
-    val mailboxesUrl = getMailboxesUrl(username)
-    wsClient.url(s"$mailboxesUrl/Outbox").put("")
+    createMailbox(username, "Outbox")
   }
 
   def createSentBox(username: Username) = {
+    createMailbox(username, "Sent")
+  }
+
+  def createMailbox(username: Username, name: String) = {
```
mbaechler: what about declaring two argument lists so that you can do partial application ?

blackheaven: Like that?

mbaechler: I was actually thinking of the opposite order so that you define the user out of the loop.

PR #53, opened 2019-03-19
Migration to Gatling 3

### LINKED ISSUE #52: Migration to Gatling 3
The goal of this issue is to make the current code compatible with Gatling 3 instead of Gatling 2.

rouazana:
Fixes #52


project/plugins.sbt
```diff
@@ -1,4 +1,4 @@
 logLevel := Level.Warn
 
-addSbtPlugin("io.gatling" % "gatling-sbt" % "2.2.0")
-addSbtPlugin("com.typesafe.sbteclipse" % "sbteclipse-plugin" % "4.0.0")
+addSbtPlugin("io.gatling" % "gatling-sbt" % "3.0.0")
+//addSbtPlugin("com.typesafe.sbteclipse" % "sbteclipse-plugin" % "4.0.0")
```
mbaechler: why?

src/main/scala-2.12/org/apache/james/gatling/smtp/SmtpProtocolBuilder.scala
```diff
@@ -1,25 +1,19 @@
...
-
 }
 
 case class SmtpProtocolBuilder(protocol: SmtpProtocol) {
 
   def host(host: String) = copy(protocol.copy(host = host))
   def ssl(ssl: Boolean) = copy(protocol.copy(ssl = ssl))
   def port(port: Int) = copy(protocol.copy(port = port))
-  def port(port: String) = if (port != null) copy(protocol.copy(port = Integer.valueOf(port))) else this
+  def port(port: String) = Option(port).map(_ -> copy(protocol.copy(port = Integer.valueOf(port)))).getOrElse(this)
```
mbaechler: does it even make sense to support null values? I mean, Scala devs don't use null ...

src/main/scala-2.12/org/apache/james/gatling/smtp/SmtpProtocolBuilder.scala
```diff
@@ -1,25 +1,19 @@
...
   def auth(auth: Boolean) = copy(protocol.copy(auth = auth))
 
   def build() = {
-    if (protocol.port == null) {
-      port(SmtpProtocol.defaultPort(protocol.ssl)).protocol
-    }
-
-    protocol
+    Option(protocol.port).map(_ => protocol).getOrElse(port(SmtpProtocol.defaultPort(protocol.ssl)).protocol)
```
mbaechler: same question here

src/test/scala-2.12/org/apache/james/gatling/jmap/scenari/FeederJmapAllScenario.scala
```diff
@@ -6,16 +6,15 @@ import org.apache.james.gatling.jmap.scenari.common.Configuration._
...
 import scala.concurrent.duration._
 
 class FeederJmapAllScenario extends Simulation {
   private val loopVariableName = "any"
 
-  val users = csv("users.csv").circular.records
-    .map({ record => User(Username(record("username")), Password(record("password"))) })
-    .map(Future.apply(_))
+  val users = csv("users.csv").circular.readRecords
+    .map({ record =>    User(Username(record("username").toString), Password(record("password").toString)) })
```
mbaechler: why these spaces ?
could we avoid calling toString on records?

src/test/scala-2.12/org/apache/james/gatling/jmap/scenari/JmapCountMailboxesScenario.scala
```diff
@@ -1,23 +1,22 @@
 package org.apache.james.gatling.jmap.scenari
 
-import io.gatling.core.Predef.{Simulation, atOnceUsers, nothingFor, rampUsers, scenario}
+import io.gatling.core.Predef.{Simulation, atOnceUsers, scenario, _}
```
mbaechler: why?

blackheaven (approved):

project/plugins.sbt
rouazana: this should be added in your personal plugin (under `.sbt/1.0/plugins/plugins.sbt`)
I will remove this line in a following fixup

chibenwa (commented):
Read it

src/main/scala-2.12/org/apache/james/gatling/smtp/SmtpProtocolBuilder.scala
remk: The protocol use a java Integer, not a scala Int. I think we should refactor it. And then we could drop the null handling.

mbaechler: Int can be null, like Integer. And we don't use null in Java neither so the refactoring, while still fine, is not a prerequisite, don't you think?

remk: Yes i agree with you.

src/test/scala-2.12/org/apache/james/gatling/jmap/scenari/JmapCountMailboxesScenario.scala
rouazana: we are using `import io.gatling.core.Predef._` everywhere but here. It's seems legit as it is a DSL. Is it ok for you if I replace this line by `import io.gatling.core.Predef._`?

mbaechler: yes

project/plugins.sbt
rouazana: removed

src/main/scala-2.12/org/apache/james/gatling/smtp/SmtpProtocolBuilder.scala
rouazana: I applied Remi's fixup about this, removing entirely the useless SmtpProtocolBuilder class

src/main/scala-2.12/org/apache/james/gatling/smtp/SmtpProtocolBuilder.scala
rouazana: removed

src/test/scala-2.12/org/apache/james/gatling/jmap/scenari/FeederJmapAllScenario.scala
rouazana: I tried something, please review

blackheaven (approved):

src/test/scala-2.12/org/apache/james/gatling/jmap/scenari/JmapCountMailboxesScenario.scala
rouazana: fixed everywhere

mbaechler (approved):

src/test/scala-2.12/org/apache/james/gatling/jmap/scenari/FeederJmapAllScenario.scala
remk: a litte nitpicking but you could transform  the Username and Password class to value class in order to avoid boxing.
You just have to make them extend AnyVal.

rouazana: not really related to this PR, but done :)

mbaechler (approved):

aduprat (approved):

PR #59, opened 2019-04-26
GITLAB-1156 [IMAP] Listing bodystructure of 1000 messages should take less than 10 seconds (f: 0.1 / 50)

blackheaven:


Arsnael:
Read it

src/main/scala-2.12/org/apache/james/gatling/imap/scenari/ImapListMessagesBodyStructureScenario.scala
```diff
@@ -0,0 +1,25 @@
...
+import io.gatling.core.Predef._
+import io.gatling.core.feeder.FeederBuilder
+import io.gatling.core.structure.ScenarioBuilder
+
+import scala.concurrent.duration._
+
+class ImapListMessagesBodyStructureScenario {
+
+  val listMessagesBodyStructure = exec(imap("select").select("INBOX").check(ok))
+    .exec(imap("fetch").fetch(MessageRanges(From(1)), AttributeList("BODY")).check(ok))
```
mbaechler: why did you choose to use `BODY` in place of `BODYSTRUCTURE`?

src/main/scala-2.12/org/apache/james/gatling/imap/scenari/ImapListMessagesBodyStructureScenario.scala
```diff
@@ -7,19 +7,21 @@ import com.linagora.gatling.imap.protocol.command.MessageRanges
...
 
-  def generate(feeder: FeederBuilder): ScenarioBuilder =
+  def generate(feeder: FeederBuilder, mailboxes: List[MailboxName]): ScenarioBuilder =
     scenario("ImapListMessagesBodyStructureScenario")
       .feed(feeder)
       .pause(1.second)
       .exec(imap("Connect").connect()).exitHereIfFailed
       .exec(imap("login").login("${username}", "${password}").check(ok))
-      .exec(listMessagesBodyStructure)
+      .exec(foreach(mailboxes.map(mailboxName => mailboxName.name), "mailboxName") {
```
mbaechler: I don't think that's what we expect from the scenario. I would say we select a random folder.

src/main/scala-2.12/org/apache/james/gatling/imap/scenari/ImapListMessagesBodyStructureScenario.scala
mbaechler: We should fetch explicitly 1000 messages and maybe ensure they are not contiguous by selecting arbitrary range

blackheaven: The issue is that in our stress test provisioning mailboxes have 260-320 messages per mailbox.

I'm a bit stuck, do you have any strategy for this?

src/test/scala-2.12/org/apache/james/gatling/simulation/imap/ImapListMessagesBodyStructureSimulation.scala
```diff
@@ -0,0 +1,24 @@
...
+import org.apache.james.gatling.imap.scenari.ImapListMessagesBodyStructureScenario
+import org.apache.james.gatling.simulation.{Configuration, HttpSettings}
+
+import scala.concurrent.duration._
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.{Await, Future}
+
+class ImapListMessagesBodyStructureSimulation extends Simulation {
+
+  private val users = new UserCreator(Configuration.BaseJamesWebAdministrationUrl).createUsersWithInboxAndOutbox(Configuration.UserCount)
```
remk: You don't want to create users, you just want to use the existing one already created in the provisioned data.
You can use the users in SimulationOnMailCorpus like it's done on the JMAP simulations

src/it/scala-2.12/org/apache/james/gatling/JamesServer.scala
```diff
@@ -36,17 +37,44 @@ object JamesServer {
...
+      putMessage(from)(to)(INBOX)
+
+    def createMailbox(username: Username)(mailboxName: MailboxName): Future[StandaloneWSRequest#Response] =
+      administration.createMailbox(username, mailboxName)
+
+
+    def addMessage(user: User): User => MailboxName => Unit =
+      putMessage(MAIL_AUTHOR)
+
+    def putMessage(from: Username)(to: User)(mailboxName: MailboxName): Unit = {
```
remk: private ?

src/it/scala-2.12/org/apache/james/gatling/JamesServer.scala
```diff
@@ -36,17 +37,44 @@ object JamesServer {
...
+      val store = session.getStore("imap")
+      store.connect(s"localhost", mappedImapPort, to.username.value, to.password.value)
+
+      // Get default folder
+      val folder = store.getFolder(mailboxName.name)
+      val message = makeMessage(from.value, to.username.value, MAIL_SUBJECT, MAIL_CONTENT)
+      folder.appendMessages(Array(new MimeMessage(session, message)))
+    }
+
+    def makeMessage(author: String, recipient: String, subject: String, content: String) = {
```
remk: private ?

src/main/scala-2.12/org/apache/james/gatling/imap/scenari/ImapListMessagesBodyStructureScenario.scala
```diff
@@ -0,0 +1,32 @@
...
+  def generate(feeder: FeederBuilder, mailboxes: Iterable[MailboxName], mailboxesListed: Int): ScenarioBuilder =
+    scenario("ImapListMessagesBodyStructureScenario")
+      .feed(feeder)
+      .pause(1.second)
+      .exec(imap("Connect").connect()).exitHereIfFailed
+      .exec(imap("login").login("${username}", "${password}").check(ok))
+      .exec{session =>
+        session.set("mailboxes", random.shuffle(mailboxes.map(_.name)).take(mailboxesListed))
+      }
+      .exec(foreach("${mailboxes}", "mailboxName") {
```
remk: You may create a group for the part you want to assert the execution time.
Then in the simulation you will be able to add assertion for the executime of this group.
You can see example of this in the JMAP scenario and simulations on the mail corpus

trantienduchn:
Read it

src/it/scala-2.12/org/apache/james/gatling/ImapListMessagesBodyStructureIT.scala
```diff
@@ -0,0 +1,25 @@
...
+import scala.concurrent.{Await, Future}
+
+
+class ImapListMessagesBodyStructureIT extends ImapIT {
+  private val MAILBOXES_LISTED = 4
+  private val mailboxes = List(MailboxName("INBOX"), MailboxName("rmbx0"), MailboxName("rmbx1.smbx5"), MailboxName("rmbx9"))
+
+  before {
+    users.foreach(server.sendMessage(Fixture.homer.username))
+    var mailboxCreations = users.flatMap(user => mailboxes.map(server.createMailbox(user.username)))
```
mbaechler: var ?!

src/it/scala-2.12/org/apache/james/gatling/JamesServer.scala
```diff
@@ -36,17 +37,44 @@ object JamesServer {
...
+    private def makeMessage(author: String, recipient: String, subject: String, content: String) = {
+      import java.io.ByteArrayInputStream
+      import java.nio.charset.Charset
+      val body =
+        s"""From: $author
+           |To: $recipient
+           |Subject: $subject
+           |
+           |$content""".stripMargin
+      new ByteArrayInputStream(body.getBytes(Charset.forName("UTF-8")))
```
mbaechler: you should use StandardCharsets instead

src/main/scala-2.12/org/apache/james/gatling/imap/scenari/ImapListMessagesBodyStructureScenario.scala
mbaechler: I suggest you grab the `msn` range returned by the SELECT operation and create an arbitrary sample of the `msn`  (say with 100 messages for now). We should than complete the operation in 1/10th of the original timing, so let's say 1 second.

src/main/scala-2.12/org/apache/james/gatling/imap/scenari/ImapListMessagesBodyStructureScenario.scala
```diff
@@ -2,33 +2,33 @@ package org.apache.james.gatling.imap.scenari
...
-          exec(listMessagesBodyStructure)
-        })
+      .exec(_.set("mailboxes", mailboxes.map(_.name)))
+      .group(ListMessageBodyStructure.name) {
+        exec(
+          exec(imap("select")
+            .select("${mailboxes.random()}")
+            .check(ok))
+          .exec(imap("fetch")
+            .fetch(MessageRanges(Range(1, mailsFetched)), AttributeList("BODYSTRUCTURE"))
```
mbaechler, line 30: I'd rather take some non-contiguous ranges but if you find it to be overkill, I can live with that version

src/it/scala-2.12/org/apache/james/gatling/ImapListMessagesBodyStructureIT.scala
```diff
@@ -0,0 +1,25 @@
...
+import org.apache.james.gatling.jmap.MailboxName
+
+import scala.concurrent.duration._
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.{Await, Future}
+
+
+class ImapListMessagesBodyStructureIT extends ImapIT {
+  private val MAILS_FETCHED = 1
+  private val mailboxes = List(MailboxName("INBOX"), MailboxName("rmbx0"), MailboxName("rmbx1.smbx5"), MailboxName("rmbx9"))
```
rouazana, line 13: what are these magic numbers?

blackheaven: Magic numbers?

rouazana: what are these 0, 1.5, 9? where do they come from?

blackheaven: I was trying to match a bit the provisioned mailboxes name in these fixtures

Arsnael (approved):

src/it/scala-2.12/org/apache/james/gatling/ImapListMessagesBodyStructureIT.scala
rouazana: if we restart the provisionning, will these names change?

blackheaven: If we do not changed the provisioning strategy, no.

Anyway, we are in integration tests here, we just want to check that it works with arbitrary mailbox names.

rouazana (approved):

remk (approved):

PR #68, opened 2019-06-13
Use a feeder for recipients and emails, instead of always sending the same email to the same recipient

rouazana:
Done by extracting mutable data to feeders.


src/main/scala-2.12/org/apache/james/gatling/jmap/scenari/JmapBigSetScenario.scala
```diff
@@ -16,12 +16,13 @@ import scala.concurrent.duration.{Duration, DurationInt}
...
 class JmapBigSetScenario {
 
-  def generate(duration: Duration, numberOfMailboxes: Int, numberOfMessages: Int, userFeeder: UserFeeder, userPicker: UserPicker): ScenarioBuilder = {
+  def generate(duration: Duration, numberOfMailboxes: Int, numberOfMessages: Int, userFeeder: UserFeeder, recipientFeeder: FeederBuilder): ScenarioBuilder = {
     def numberOfMailboxesPerUser: Int = numberOfMailboxes + numberOfSystemMailboxes
 
     scenario("JMAP scenario on multiple mailboxes containing multiple messages")
       .feed(userFeeder)
-      .exec(CommonSteps.provisionUsersWithMailboxesAndMessages(userPicker, numberOfMailboxes, numberOfMessages))
+      .feed(recipientFeeder)
```
mbaechler: I find it a bit weird to use a feeder for something that is related to a specific action.
I would rather inject a random user in the session at the right time by doing:
```scala
.exec(session => session.put("recipient", userPicker.pick()))
```

src/test/scala-2.12/org/apache/james/gatling/simulation/jmap/FeederJmapAllSimulation.scala
```diff
@@ -3,7 +3,7 @@ package org.apache.james.gatling.simulation.jmap
 import io.gatling.core.Predef._
 import io.gatling.core.scenario.Simulation
 import org.apache.james.gatling.control.{Password, RandomUserPicker, User, UserFeeder, Username}
-import org.apache.james.gatling.jmap.scenari.FeederJmapAllScenario
+import org.apache.james.gatling.jmap.scenari.JmapAllScenario
 import org.apache.james.gatling.simulation.{Configuration, HttpSettings}
 
 class FeederJmapAllSimulation extends Simulation {
```
remk: there is already a simulation for this scenario JmapAllSimulation, you can drop the simulation too

src/main/scala-2.12/org/apache/james/gatling/jmap/scenari/JmapBigSetScenario.scala
rouazana: for me this is exactly what a feeder is doing. The UserPicker reproduces manually what a feeder is doing (taking a random user from a Seq), but putting it in the session must be done apart.

If you prefer I can use this feeder only in the sendMessages method, as I did for the MailFeeder.

src/it/scala-2.12/org/apache/james/gatling/JmapIT.scala
```diff
@@ -31,9 +31,10 @@ abstract class JmapIT extends GatlingFunSpec {
     server.stop()
   }
 
-  protected def scenario(scenarioFromFeeder: FeederBuilder => ScenarioBuilder) = {
+  protected def scenario(scenarioFromFeeder: (FeederBuilder, FeederBuilder) => ScenarioBuilder) = {
     val feeder = UserFeeder.toFeeder(users)
-    scenarioFromFeeder(feeder).actionBuilders.reverse.foreach { actionBuilder =>
+    val recipientFeeder = RecipientFeeder.userstoFeeder(users)
```
remk: s/userstoFeeder/usersToFeeder

src/it/scala-2.12/org/apache/james/gatling/JmapMessageUpdateIT.scala
```diff
@@ -11,7 +11,7 @@ class JmapMessageUpdateIT extends JmapIT {
     users.foreach(server.sendMessage(Fixture.homer.username))
   }
 
-  scenario(feederBuilder => {
-    new JmapMessageUpdateScenario().generate(10 seconds, UserFeeder.toFeeder(users), RandomUserPicker(users), 10)
+  scenario((feederBuilder, recipientFeederBuilder) => {
+    new JmapMessageUpdateScenario().generate(10 seconds, UserFeeder.toFeeder(users), recipientFeederBuilder, 10)
```
remk: sorry it's not you, but what is the 10 for ?

src/main/scala-2.12/org/apache/james/gatling/control/RecipientFeeder.scala
```diff
@@ -0,0 +1,24 @@
+package org.apache.james.gatling.control
+
+import org.apache.james.gatling.jmap.RecipientAddress
+
+object RecipientFeeder {
+
+  type RecipientFeeder = Array[Map[String, String]]
+
+  val RecipientSessionParam = "recipient"
```
remk: RECIPIENT_SESSION_PARAM
i'm ok with recipientSessionParam too.
But here we mistake it with a type name

src/main/scala-2.12/org/apache/james/gatling/jmap/scenari/JmapBigSetScenario.scala
remk: I agree that your solution feel simpler.

src/main/scala-2.12/org/apache/james/gatling/jmap/JmapChecks.scala
```diff
@@ -6,6 +6,6 @@ object JmapChecks {
 
   val noError = jsonPath("$.error").notExists
 
-  def created(messageId: MessageId) = jsonPath(s"$$[0][1].created['${messageId.id}'].id").exists
```
remk, line 9: remove the 's' for interpolating, then you will be able to use only one '$' and remove the curly braces surrounding 'messageId'

src/main/scala-2.12/org/apache/james/gatling/jmap/JmapMessages.scala
```diff
@@ -29,18 +29,18 @@ object JmapMessages {
...
-  def sendMessages(messageId: MessageId, subject: Subject, textBody: TextBody) =
+  def sendMessages() =
     JmapAuthentication.authenticatedQuery("sendMessages", "/jmap")
       .body(StringBody(
         s"""[[
           "setMessages",
           {
             "create": {
-              "${messageId.id}" : {
+              "$${messageId}" : {
```
remk: idem you could remove the scala string interpolation and simplify all this string

src/main/scala-2.12/org/apache/james/gatling/jmap/JmapChecks.scala
```diff
@@ -6,6 +6,6 @@ object JmapChecks {
 
   val noError = jsonPath("$.error").notExists
 
-  def created() = jsonPath(s"$$[0][1].created['$${messageId}'].id").exists
```
remk: _

src/main/scala-2.12/org/apache/james/gatling/jmap/JmapMailbox.scala
```diff
@@ -36,6 +36,7 @@ object JmapMailbox {
 
   implicit val reads = Json.reads[JmapMailbox]
 
+  val OutboxMailboxIdSessionParam = "outboxMailboxId"
```
remk: no need to capitalize, it's not a type.

src/main/scala-2.12/org/apache/james/gatling/jmap/JmapMessages.scala
```diff
@@ -29,19 +29,23 @@ object JmapMessages {
   type JmapParameters = Map[String, Any]
   val NO_PARAMETERS : JmapParameters = Map()
 
+  val MessageIdSessionParam = "messageId"
+  val SubjectSessionParam = "subject"
+  val TextBodySessionParam = "textBody"
```
remk: no need to capitalize, it's not a type.

uppercase snake case or lowercase camel case are ok

rouazana: you are right, but I took this convention from UserFeeder. If the last commit makes consensus, I will update every constant

src/main/scala-2.12/org/apache/james/gatling/jmap/JmapChecks.scala
Arsnael: Just curious, what does the double `$$` stands for in a string interpolation? Cannot really find doc about it sorry

Arsnael:
Read it

src/main/scala-2.12/org/apache/james/gatling/jmap/JmapChecks.scala
rouazana: It's to escape the `$` sign in a string interpolation (string starting by `s"`). See https://www.scala-lang.org/files/archive/spec/2.13/01-lexical-syntax.html#string-literals

src/main/scala-2.12/org/apache/james/gatling/jmap/scenari/JmapBigSetScenario.scala
rouazana: I moved the feeder. At the end it just replaces the UserPicker with proper Gatling meaning. I also removed the UserPicker by the way.

blackheaven (approved):

src/it/scala-2.12/org/apache/james/gatling/JmapIT.scala
```diff
@@ -31,9 +31,10 @@ abstract class JmapIT extends GatlingFunSpec {
     server.stop()
   }
 
-  protected def scenario(scenarioFromFeeder: FeederBuilder => ScenarioBuilder) = {
+  protected def scenario(scenarioFromFeeder: (FeederBuilder, FeederBuilder) => ScenarioBuilder) = {
```
mbaechler: I don't like to rely on the fact people know what the two expected feeders are. Could you define types?
Like UserFeeder(Builder) & RecipientsFeeder(Builder)?

remk (approved):

src/it/scala-2.12/org/apache/james/gatling/JmapIT.scala
rouazana: fixed

blackheaven (approved):

Arsnael (approved):

remk (approved):

chibenwa (approved):

mbaechler (approved):

PR #70, opened 2019-06-25
Drop duration from org.apache.james.gatling.simulation.imap.ImapFullSimulation

blackheaven:


src/main/scala-2.12/org/apache/james/gatling/imap/scenari/ImapFullScenario.scala
```diff
@@ -30,21 +30,11 @@ class ImapFullScenario {
...
-      .during(duration) {
-        randomSwitch(
-          75.0 -> exec(lightUser),
-          25.0 -> exec(heavyUser)
-        )
-      }
-}
\ No newline at end of file
+      .exec(receiveEmail
+        .exec(readLastEmail))
```
rouazana: you don't loop at all?

rouazana: why not
```
.exec(receiveEmail)
.exec(readLastEmail)
```
 ?

rouazana: why did you remove different usages (light / heavy)?

remk: I think the best way would be to inject two different profiles in the simulation the light users and the heavy ones.
The light one would play a ImapFullLightScenario and the heavy ones a ImapFullHeavyScenario.

What do you think of that ?

trantienduchn (commented):
read it

src/main/scala-2.12/org/apache/james/gatling/imap/scenari/ImapFullScenario.scala
Arsnael: I think he removed light and heavy users to drop the duration of the ImapFullScenario, as in the title... I'm not sure though to understand the point of that?

rouazana: nice solution, this will better balance the scenarios

remk: > I think he removed light and heavy users to drop the duration of the ImapFullScenario, as in the title... I'm not sure though to understand the point of that?

Yes, we should drop the duration in the scenario. My proposal is, in addition to that to drop the switch in the scenario. Make 2 different scenario for each usage. And in the simulation inject 75% of users with the light scenario and 25% with the heavy one. 
I hope its clearer.


And the point to remove the duration in the scenario is that the duration is already defined in the simulation. Having a duration defined in the scenario for each user injected make the simulation taking much more time that what is expected.

Arsnael: Oh ok ! Thanks for the clarification @remk , still not familiar enough with the gatling simulations and scenarios :)

mbaechler (commented):
I describe a solution with two kinds of users but this PR does not implement the solution. Couldn't it be done in this PR?

rouazana (approved):

remk (approved):

Arsnael (approved):

src/main/scala-2.12/org/apache/james/gatling/imap/scenari/ImapFullHeavyUserScenario.scala
```diff
@@ -0,0 +1,41 @@
...
+import com.linagora.gatling.imap.protocol.command.MessageRanges
+import io.gatling.core.Predef._
+import io.gatling.core.feeder.FeederBuilder
+import io.gatling.core.structure.ScenarioBuilder
+
+import scala.concurrent.duration._
+
+class ImapFullHeavyUserScenario {
+
+  val receiveEmail = exec(imap("append").append("INBOX", Some(scala.collection.immutable.Seq("\\Flagged")), Option.empty[Calendar],
```
mbaechler: I agree to duplicate the whole generate method but not the two vals

Arsnael (approved):

chibenwa (approved):

mbaechler (approved):

Issue #69, opened 2019-06-21
Fix JMAP getMessages behaviour


rouazana:
getMessages has currently two related issues:
- first defining properties like `"properties": [ ${Json.stringify(properties)} ]` make a json array of a json array, instead of a json array. Consequence: the JMAP request cannot be parsed, and the server errors it
- second JmapChecks.noError does not detect the error returned in this case, which is nevertheless a perfectly valid JMAP error `[["error", etc.]]`

PR #71, opened 2019-06-25
James-gatling 69 :  fix jmap get messages

remk:
getMessages has currently two related issues:

    first defining properties like "properties": [ ${Json.stringify(properties)} ] make a json array of a json array, instead of a json array. Consequence: the JMAP request cannot be parsed, and the server errors it
    second JmapChecks.noError does not detect the error returned in this case, which is nevertheless a perfectly valid JMAP error [["error", etc.]]

https://github.com/linagora/james-gatling/issues/69

### DIFF
--- a/src/main/scala-2.12/org/apache/james/gatling/jmap/JmapChecks.scala
+++ b/src/main/scala-2.12/org/apache/james/gatling/jmap/JmapChecks.scala
@@ -1,10 +1,14 @@
 package org.apache.james.gatling.jmap
 
 import io.gatling.core.Predef._
+import io.gatling.core.check.extractor.jsonpath.{JsonPathCheckBuilder, JsonPathOfType}
 
 object JmapChecks {
 
-  val noError = jsonPath("$.error").notExists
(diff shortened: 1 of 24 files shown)


src/main/scala-2.12/org/apache/james/gatling/jmap/JmapChecks.scala
```diff
@@ -4,7 +4,7 @@ import io.gatling.core.Predef._
 
 object JmapChecks {
 
-  val noError = jsonPath("$.error").notExists
+  val noError = jsonPath("$[?(@[0] == 'error')]").notExists
```
rouazana: I would really like a test to check that it's working

trantienduchn (commented):
read it

chibenwa (approved):

Arsnael (approved):

rouazana (approved):

Arsnael (approved):

blackheaven (approved):

chibenwa (approved):

PR #76, opened 2020-02-27
Add mixed body size SMTP Simulation

blackheaven:

PR #78, opened 2020-03-13
Open injection fixup

Arsnael:
So before #76 changes, for the simulations injecting a fixed amount of users per second, we had around 13 (I did the math :) ).

With default values of 50000 users and a duration of one hour, and one hour being 3600 seconds we were doing `50000 / 3600 = 13.8 users injected per seconds`

Well after the update it became `50000 * 3600 / 3600 = 50000 users injected per seconds`

No wonder charge-01 was out of memory in less than a minute when I tried to run those :trollface:

### DIFF
--- a/src/test/scala-2.12/org/apache/james/gatling/simulation/Injection.scala
+++ b/src/test/scala-2.12/org/apache/james/gatling/simulation/Injection.scala
@@ -9,8 +9,8 @@ abstract sealed class UsersDensity {
   def injectDuring(givenDuring: FiniteDuration): OpenInjectionStep
 }
 case class UsersPerHour(nb: Double) extends UsersDensity {
-  private def usersPerSecForDuration(givenDuring: FiniteDuration): Double = nb * givenDuring.toSeconds / 1.hour.toSeconds
-  override def injectDuring(givenDuring: FiniteDuration): ConstantRateOpenInjection = constantUsersPerSec(usersPerSecForDuration(givenDuring)) during givenDuring
+  private def usersPerSecForDuration(): Double = nb / 1.hour.toSeconds
+  override def injectDuring(givenDuring: FiniteDuration): ConstantRateOpenInjection = constantUsersPerSec(usersPerSecForDuration()) during givenDuring
 }
 case class UsersTotal(nb: Double) extends UsersDensity {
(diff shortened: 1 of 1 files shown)


chibenwa (approved):

src/test/scala-2.12/org/apache/james/gatling/simulation/Injection.scala
```diff
@@ -9,8 +9,8 @@ abstract sealed class UsersDensity {
   def injectDuring(givenDuring: FiniteDuration): OpenInjectionStep
 }
 case class UsersPerHour(nb: Double) extends UsersDensity {
-  private def usersPerSecForDuration(givenDuring: FiniteDuration): Double = nb * givenDuring.toSeconds / 1.hour.toSeconds
-  override def injectDuring(givenDuring: FiniteDuration): ConstantRateOpenInjection = constantUsersPerSec(usersPerSecForDuration(givenDuring)) during givenDuring
+  private def usersPerSecForDuration(): Double = nb / 1.hour.toSeconds
```
blackheaven: Then, you will inject the same amount of user whatever the duration is?

Arsnael: It depends what you want here exactly... I put it back to what it was before, because anyway the previous math was wrong

Maybe it should be more then `nb / givenDuring.toSeconds()` then?

Number of users I want to inject in the total period of time?

blackheaven: I've just tried it, I think it's good, just drop the `()`

blackheaven (approved):

src/test/scala-2.12/org/apache/james/gatling/simulation/Injection.scala
Arsnael: actually I rechanged it to `nb / givenDuring.toSeconds`

the function name speaks for itself. We want the uers per sec for duration. `nb` is the total number of users we want to inject for the duration of the simulation, `givenDuring` is the total duration of the simulation

So users injected per second for the total duration should be `nb / givenDuring.toSeconds`, it's more logical this way.

chibenwa (approved):

blackheaven (approved):

Arsnael:
Merged

PR #80, opened 2020-06-05
Add JmapReadOnlyScenario

blackheaven:


src/main/scala-2.12/org/apache/james/gatling/jmap/JmapMessages.scala
```diff
@@ -150,7 +152,13 @@ object JmapMessages {
   val listMessagesChecks: Seq[HttpCheck] = List(
       status.is(200),
       JmapChecks.noError,
-      jsonPath("$[0][1].messageIds[*]").findAll.saveAs("messageIds"))
```
chibenwa, line 153: :+1: 

Maybe we could add a seperate check for 'non emptyness' that existing scenario could compose in order not to change their behaviour?

src/main/scala-2.12/org/apache/james/gatling/jmap/scenari/JmapReadOnlyScenario.scala
```diff
@@ -0,0 +1,80 @@
...
+  val poll : ChainBuilder = exec(Queries.getMessagesList)
+    .exec((session: Session) => session.set("poll", pollTimeout.fromNow))
+  val shouldPoll: Expression[Boolean] = (session: Session) => {
+    session.contains("poll") && session("poll").as[Deadline].timeLeft <= safetyMargin
+  }
+
+  def generate(userFeeder: AuthenticatedUserFeederBuilder, duration: Duration): ScenarioBuilder = {
+    scenario("JmapReadOnlyScenario")
+      .feed(userFeeder)
+      .during(duration) {
```
chibenwa, line 69: Maybe we should add a pause to add some temporisation (like interact with the mail server every 30s)?

chibenwa (approved):
Looks good, we just remain the JWT part, right?

src/test/scala-2.12/org/apache/james/gatling/simulation/JmapReadOnlySimulation.scala
```diff
@@ -0,0 +1,30 @@
...
+import io.gatling.core.scenario.Simulation
+
+class JmapReadOnlySimulation extends Simulation {
+
+  private def recordValueToString(recordValue: Any): String = recordValue match {
+    case s: String => s
+    case a: Any => println("Warning: calling toString on a feeder value"); a.toString
+  }
+
+  val authenticatedUsers: Iterator[AuthenticatedUser] = csv("usersAuthenticated.csv").readRecords
```
Arsnael, line 17: Where is that csv file?

chibenwa: As it may contain some real users we should not commit it here o avoid some information disclosure, I guess. Maybe we could provide an exemple?

src/main/scala-2.12/org/apache/james/gatling/control/UserFeeders.scala
```diff
@@ -19,3 +19,20 @@ object UserFeeder {
...
 }
+
+object AuthenticatedUserFeeder {
+
+  type AuthenticatedUserFeeder = Iterator[Map[String, String]]
+
+  type AuthenticatedUserFeederBuilder = () => Feeder[Any]
+
+  val usernameSessionParam = "username"
+  val accessTokenSessionParam = "accessToken"
```
chibenwa: Can we disambiguate if it is a JMAP access token or a JWT one?

src/test/scala-2.12/org/apache/james/gatling/simulation/JmapReadOnlySimulation.scala
Arsnael: Was not sure if it would be real data or not :)

src/main/scala-2.12/org/apache/james/gatling/jmap/scenari/JmapReadOnlyScenario.scala
blackheaven: I do not get it

src/test/scala-2.12/org/apache/james/gatling/simulation/JmapReadOnlySimulation.scala
blackheaven: I have just set the filename, no leak here.

blackheaven:
> Looks good, we just remain the JWT part, right?

filling the file csv file and we  are good to go.

chibenwa (approved):
An example CSV file in test resources would IMO be welcome, but I'm 100% ok with this work.

Koodo!

remk (approved):

PR #81, opened 2020-06-10
Multiples JmapReadOnlyScenario fixes

blackheaven:


src/test/scala-2.12/org/apache/james/gatling/simulation/jmap/JmapReadOnlySimulation.scala
```diff
@@ -15,17 +15,16 @@ class JmapReadOnlySimulation extends Simulation {
...
         username = Username(recordValueToString(record("username"))),
         jwtAccessToken = JwtAccessToken(recordValueToString(record("jwtAccessToken")))))
-    .toIterator
 
-  val feeder: AuthenticatedUserFeederBuilder = AuthenticatedUserFeeder.toFeeder(authenticatedUsers)
+  val feeder: AuthenticatedUserFeederBuilder = AuthenticatedUserFeeder.toFeeder(authenticatedUsers.toIterator)
 
   setUp(new JmapReadOnlyScenario().generate(feeder, Configuration.ScenarioDuration)
-    .inject(UsersTotal(Configuration.UserCount).injectDuring(Configuration.InjectionDuration))
+    .inject(UsersTotal(authenticatedUsers.length).injectDuring(Configuration.InjectionDuration))
```
rouazana, line 27: does it still work with a circular feeder? (but maybe it's not needed for this particular simulation)

Arsnael (approved):

src/main/scala-2.12/org/apache/james/gatling/jmap/scenari/JmapReadOnlyScenario.scala
```diff
@@ -68,7 +72,8 @@ class JmapReadOnlyScenario {
       .feed(userFeeder)
       .during(duration) {
         group(InboxHomeLoading.name)(
-          doIf(shouldPoll) {
+          pause(1 second)
```
chibenwa, line 75: We should have a 30s pause to match inbox behaviour

src/main/scala-2.12/org/apache/james/gatling/jmap/scenari/JmapReadOnlyScenario.scala
```diff
@@ -68,7 +72,8 @@ class JmapReadOnlyScenario {
...
+          .doIf(shouldPoll) {
             exec(poll)
           }
             .exec(Queries.getMailboxes)
```
chibenwa, line 79: We should not excute these queries if we had been polling. Is it the case?

src/test/scala-2.12/org/apache/james/gatling/simulation/jmap/JmapReadOnlySimulation.scala
chibenwa: Yes circular feeder don't makes sense for this one.

chibenwa (commented):
:+1: 

(thanks)

chibenwa (approved):

rouazana (approved):

Arsnael (approved):

PR #82, opened 2020-09-25
JAMES-3028 Fix the constant number of users being injected for corpus…

Arsnael:
… simulations and the injection duration

We had Gatling simulation trying to inject around 166 users per second for a duration of just 5 minutes... This was unrealistic.

This will allow to inject instead around 14 users per second for a duration of an hour.

Not the first time I fix this rate BTW...

### DIFF
--- a/src/test/scala-2.12/org/apache/james/gatling/simulation/Injection.scala
+++ b/src/test/scala-2.12/org/apache/james/gatling/simulation/Injection.scala
@@ -10,8 +10,8 @@ abstract sealed class UsersDensity {
 }
 case class UsersPerHour(nb: Double) extends UsersDensity {
   private val ONE_HOUR: Duration = 1 hour
-  private def usersPerSecForDuration: Double = nb / ONE_HOUR.toSeconds
-  override def injectDuring(givenDuring: FiniteDuration): ConstantRateOpenInjection = constantUsersPerSec(usersPerSecForDuration) during givenDuring
+  private val usersPerSec: Double = nb / ONE_HOUR.toSeconds
+  override def injectDuring(givenDuring: FiniteDuration): ConstantRateOpenInjection = constantUsersPerSec(usersPerSec) during givenDuring
 }
 case class UsersPerSecond(nb: Double) extends UsersDensity {
   override def injectDuring(givenDuring: FiniteDuration): ConstantRateOpenInjection = constantUsersPerSec(nb) during givenDuring
--- a/src/test/scala-2.12/org/apache/james/gatling/simulation/Injection.scala
+++ b/src/test/scala-2.12/org/apache/james/gatling/simulation/Injection.scala
@@ -8,9 +8,13 @@ import scala.concurrent.duration._
 abstract sealed class UsersDensity {
   def injectDuring(givenDuring: FiniteDuration): OpenInjectionStep
 }
+case class UsersPerHour(nb: Double) extends UsersDensity {
+  private val ONE_HOUR: Duration = 1 hour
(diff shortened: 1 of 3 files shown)


src/test/scala-2.12/org/apache/james/gatling/simulation/SimulationOnMailCorpus.scala
```diff
@@ -27,7 +27,7 @@ trait SimulationOnMailCorpus {
 
   protected val feeder: SourceFeederBuilder[String] = UserFeeder.toFeeder(getUsers).circular
 
-  protected def injectUsersInScenario(scenario: ScenarioBuilder, nbUsers: UsersDensity = UsersPerHour(50000)) = {
+  protected def injectUsersInScenario(scenario: ScenarioBuilder, nbUsers: UsersDensity = UsersPerSecond(50000)) = {
```
chibenwa: 50.000 user per second? 

I understand it is a default and that simulation specify their own value but maybe that should be lowered....

Arsnael: No I just tried to rename it and obviously the renaming is wrong :)

What it does it takes the number of users / duration in seconds... I put back the default duration at 1 hour so it goes to around 13 users per second 

I'm not sure how should I name this class... UsersPerHour is wrong (UsersPerSecond is not better)

src/test/scala-2.12/org/apache/james/gatling/simulation/Injection.scala
```diff
@@ -8,7 +8,7 @@ import scala.concurrent.duration._
 abstract sealed class UsersDensity {
   def injectDuring(givenDuring: FiniteDuration): OpenInjectionStep
 }
-case class UsersPerHour(nb: Double) extends UsersDensity {
+case class UsersPerSecond(nb: Double) extends UsersDensity {
   private def usersPerSecForDuration(givenDuring: FiniteDuration): Double = nb / givenDuring.toSeconds
   override def injectDuring(givenDuring: FiniteDuration): ConstantRateOpenInjection = constantUsersPerSec(usersPerSecForDuration(givenDuring)) during givenDuring
```
chibenwa: This should obviously become : 

```
override def injectDuring(givenDuring: FiniteDuration): ConstantRateOpenInjection = constantUsersPerSec(usersPerSecForDuration(1 hour)) during givenDuring
```

:-)

`UserPerHour(12000).injectDuring(1 hour)` and `UserPerHour(12000).injectDuring(5 minutes)` should of course inject at the same rate.

src/test/scala-2.12/org/apache/james/gatling/simulation/Injection.scala
```diff
@@ -8,9 +8,13 @@ import scala.concurrent.duration._
 abstract sealed class UsersDensity {
   def injectDuring(givenDuring: FiniteDuration): OpenInjectionStep
 }
+case class UsersPerHour(nb: Double) extends UsersDensity {
+  private val ONE_HOUR: Duration = 1 hour
+  private def usersPerSecForDuration: Double = nb / ONE_HOUR.toSeconds
```
chibenwa: val

chibenwa: s/usersPerSecForDuration/usersPerSec

Arsnael:
With provisioning, InboxHomeLoadingSimulation : 

screenshot--2020 09 28-16_54_26

chibenwa (approved):
Good job!

nvduc91 (approved):
nice!

PR #83, opened 2020-10-14
JAMES-3028 Imap perf tests fixes

Arsnael:


src/main/scala-2.12/org/apache/james/gatling/imap/scenari/ImapListMessagesBodyStructureScenario.scala
```diff
@@ -27,7 +27,7 @@ class ImapListMessagesBodyStructureScenario {
             .select("${mailboxes.random()}")
             .check(ok))
           .exec(imap("fetch")
-            .fetch(MessageRanges(Range(1, mailsFetched)), AttributeList("BODYSTRUCTURE"))
+            .fetch(MessageRanges(From(1L)), AttributeList("BODYSTRUCTURE"))
```
chibenwa: I'm reluctant for that change, as the performance depends highly on the count of fetched items...

Maybe if doable we could provision more messages?

Maybe we could decrease the amount of mail fetched?

Arsnael: I reverted the code and put a limit of 50 for mails fetched. As we saw on the provisioning by checking Cassandra, we have an average of a 100 messages per mailbox more or less 10 on it. With a limit of 50 we are pretty sure it's not gonna fail for this. 50 should be enough as well for now.

chibenwa (approved):

src/test/scala-2.12/org/apache/james/gatling/simulation/imap/ImapListMessagesBodyStructureSimulation.scala
```diff
@@ -3,15 +3,15 @@ package org.apache.james.gatling.simulation.imap
...
-
+import org.apache.james.gatling.simulation.{Configuration, HttpSettings, SimulationOnMailCorpus, UsersPerHour}
 import com.linagora.gatling.imap.PreDef.imap
 
 import scala.concurrent.duration._
 
 class ImapListMessagesBodyStructureSimulation extends Simulation with SimulationOnMailCorpus {
-  private val MAILS_FECTHED = 200
-  setUp(injectUsersInScenario(new ImapListMessagesBodyStructureScenario().generate(feeder, getMailboxes, MAILS_FECTHED)))
+  private val MAILS_FECTHED = 50
```
LanKhuat: s/FECTHED/FETCHED

src/main/scala-2.12/org/apache/james/gatling/jmap/scenari/JmapSelectArbitraryMaiboxScenario.scala
```diff
@@ -26,7 +26,7 @@ class JmapSelectArbitraryMaiboxScenario(minMessagesInMailbox: Int) {
   private val openArbitraryMailboxes: HttpRequestBuilder = getRandomMessages(openpaasInboxOpenMessageProperties, Keys.messageIds)
 
   def generate(userFeeder: UserFeederBuilder): ScenarioBuilder = {
-    scenario("JmapOpenArbitraryMessageScenario")
+    scenario("JmapSelectArbitraryMaiboxScenario")
```
LanKhuat: s/Maibox/Mailbox

LanKhuat (approved):

chibenwa (approved):

PR #94, opened 2020-10-27
Switch ImapFullSimulation to constant rate injection

chibenwa:


Arsnael:
I think that's more or less what I did when I forked it into `ImapFullWithDurationSimulation` ? (except I added a defined duration... anyway I was not sure but maybe we don't need to keep both tests here)

chibenwa:
> (except I added a defined duration... anyway I was not sure but maybe we don't need to keep both tests here)

Yes, let's remove ImapFullWithDurationSimulation

PR #96, opened 2020-10-28
Add an env parameter for the duration of the simulation

Arsnael:


src/test/scala-2.12/org/apache/james/gatling/simulation/Configuration.scala
```diff
@@ -20,8 +20,13 @@ object Configuration {
...
 
-  val ScenarioDuration = 1 hour
-  val InjectionDuration = 1 hour
+  val DURATION_PROPERTY = Properties.envOrNone("DURATION") match {
+    case Some(duration) => Some(duration.toInt minutes)
+    case _ => None
+  }
+
+  val ScenarioDuration = DURATION_PROPERTY.getOrElse(1 hour)
+  val InjectionDuration = DURATION_PROPERTY.getOrElse(1 hour)
```
Arsnael, line 29: Keeping both for now until I get this to work but wondering if should keep both... There is few differences in the code obviously but I'm still not sure exactly what was initially intended here.

chibenwa: Inject say 1.000 users over 1 minutes then play a simulation with all of them (loop) for 1 hour. 

Let's keep this.

chibenwa (approved):

LanKhuat (approved):

Arsnael:
merged

PR #103, opened 2020-11-20
JMAP-DRAFT Provide a PlatformValidationSimulation

chibenwa:


Arsnael (approved):
The refactorings could have been separated into an other commit... Otherwise looks good :)

PR #107, opened 2020-11-30
IMAP Plateform scenarii: Decrease APPEND count

chibenwa:
This is not representative of a real usage

### DIFF
--- a/src/main/scala-2.12/org/apache/james/gatling/imap/scenari/PlatformValidationScenario.scala
+++ b/src/main/scala-2.12/org/apache/james/gatling/imap/scenari/PlatformValidationScenario.scala
@@ -9,10 +9,9 @@ import scala.concurrent.duration.{Duration, _}
 
 class PlatformValidationScenario(minWaitDelay: Duration = 1 minute, maxWaitDelay: Duration = 5 minutes)  {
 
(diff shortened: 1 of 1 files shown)


remk (approved):

src/main/scala-2.12/org/apache/james/gatling/imap/scenari/PlatformValidationScenario.scala
```diff
@@ -9,10 +9,9 @@ import scala.concurrent.duration.{Duration, _}
 
 class PlatformValidationScenario(minWaitDelay: Duration = 1 minute, maxWaitDelay: Duration = 5 minutes)  {
 
-  val lightScenario: ChainBuilder = exec(ImapCommonSteps.receiveEmail)
-    .exec(ImapCommonSteps.readLastEmail)
+  val lightScenario: ChainBuilder = exec(ImapCommonSteps.readLastEmail)
```
rouazana, line 12: a light user never receive an email?

chibenwa: A light user never do APPEND.

APPEND is a terrible way to simulate email reception.

Email reception rate is very low. Maybe 100 mail a day for intensive users. That's one mail every 5 minutes, that's what the heavy simulation accounts for.

PR #108, opened 2020-12-07
Uniformize env and configuration with openpaas-gatling

rouazana:


src/test/scala-2.12/org/apache/james/gatling/simulation/Configuration.scala
```diff
@@ -20,16 +20,12 @@ object Configuration {
...
-  }
-
-  val ScenarioDuration = DURATION_PROPERTY.getOrElse(1 hour)
-  val InjectionDuration = DURATION_PROPERTY.getOrElse(1 hour)
-  val UserCount = 100
-  val RandomlySentMails = 10
-  val NumberOfMailboxes = 10
-  val NumberOfMessages = 20
+  val InjectDuration = Properties.envOrElse("INJECT_DURATION", "10").toInt seconds
+  val ScenarioDuration = Properties.envOrElse("SCENARIO_DURATION", "10").toInt seconds
```
Arsnael, line 24: That would potentially need to be synced with the bench before being merged I think? as we have a duration injected to simulate a warmup

Arsnael: And likely the stress file test in james too

chibenwa (commented):
Agree, let's make sure we do not break the bench.

Use a feature branch if you need the change quickly?

rouazana:
I'm not sure it prevents this to be merged, as nothing forces you to pull on the bench.
Anyway finally I don't need this fix urgently to have a full platform scenario, so merge it when it's better for you.

PR #109, opened 2021-01-07
Fetch full message and not only bodystructure

remk:


Arsnael (approved):
Nice catch, thanks @remk

rouazana (approved):

chibenwa:
It do not change much on the DB reads, however transmitting more info on the wire to gatling might worsen memory problems.

How do we stand after this fix regarding memory usage? Still able to run 4000 users on the plateform simulation stuff?

remk:
> It do not change much on the DB reads, however transmitting more info on the wire to gatling might worsen memory problems.
> 
> How do we stand after this fix regarding memory usage? Still able to run 4000 users on the plateform simulation stuff?

Yes, i ran the platform simulation with 5k users, it broke around 4600 by an out of memory exception in gatling.

chibenwa:
Ok :+1: then as it to not bring more limitations

PR #112, opened 2021-04-26
ISSUE-105 Add websocket support

Arsnael:
Not that easy after all... I've got a few issues with the IT test so far...

* It's a circle right in terms of user feeding? My memory is a bit rusty... So of course with only one user it opens the websocket first and tries to open new connections again (but failed because one opened already). So we can as well open multiple websocket connections per virtual user but it requires using the method `wsName`. Do we need that, or should we just open one connection per user?
* The check fails. It tells me found nothing. The problem is I'm not sure what kind of answer I get. I might need to dig more on that. Could not find a way to try print the possible response we get for debugging too. 

Those are my main issues. But at least the client connects successfully with websockets.

### DIFF
--- a/src/main/scala-2.12/org/apache/james/gatling/jmap/rfc8621/scenari/WebsocketSimpleScenario.scala
+++ b/src/main/scala-2.12/org/apache/james/gatling/jmap/rfc8621/scenari/WebsocketSimpleScenario.scala
@@ -1,0 +1,36 @@
+package org.apache.james.gatling.jmap.rfc8621.scenari
+
+import io.gatling.core.Predef._
+import io.gatling.core.structure.{ChainBuilder, ScenarioBuilder}
+import io.gatling.http.Predef._
+import org.apache.james.gatling.control.UserFeeder.UserFeederBuilder
+import org.apache.james.gatling.jmap.draft.CommonSteps.provisionSystemMailboxes
+import org.apache.james.gatling.jmap.draft.{MailboxId, MailboxName}
+import org.apache.james.gatling.jmap.rfc8621.JmapWebsocket.{enablePush, setMailboxesWs, websocketClose, websocketConnect}
+import org.apache.james.gatling.jmap.rfc8621.SessionStep
+
+import scala.concurrent.duration._
+
+class WebsocketSimpleScenario {
+  def generate(duration: Duration, userFeeder: UserFeederBuilder): ScenarioBuilder =
+    scenario("WebsocketSimpleScenario")
+      .feed(userFeeder)
+      .exec(SessionStep.retrieveAccountId)
+      .exec(provisionSystemMailboxes())
+      .exec(websocketConnect().onConnected(
+        exec(enablePush)
+          .during(duration) {
+            exec(createMailbox)
+              .pause(2 second)
+        }))
+      .exec(websocketClose())
+
+  def createMailbox: ChainBuilder =
+    exec((session: Session) => session.set("createdId", MailboxId.generate().id))
+      .exec((session: Session) => session.set("mailboxName", MailboxName.generate().name))
+      .exec(setMailboxesWs
+        .await(2 seconds)(
+          ws.checkTextMessage("check push state").check(jsonPath("$.@type").is("StateChange")))
+        .await(2 seconds)(
+          ws.checkTextMessage("check response").check(jsonPath("$.methodResponses[0][1].created").find.saveAs("mailboxCreated"))))
+}
--- a/src/it/scala-2.12/org/apache/james/gatling/jmap/draft/JmapIT.scala
+++ b/src/it/scala-2.12/org/apache/james/gatling/jmap/draft/JmapIT.scala
@@ -1,7 +1,5 @@
 package org.apache.james.gatling.jmap.draft
 
-import java.net.URL
-
 import io.gatling.core.Predef._
 import io.gatling.core.funspec.GatlingFunSpec
 import io.gatling.core.protocol.Protocol
@@ -17,6 +15,7 @@ import org.apache.james.gatling.control.{AuthenticatedUser, AuthenticatedUserFee
 import org.slf4j
 import org.slf4j.LoggerFactory
 
+import java.net.URL
 import scala.concurrent.Await
 import scala.concurrent.duration._
 
@@ -25,7 +24,9 @@ abstract class JmapIT extends GatlingFunSpec {
 
   protected val server: RunningServer = JamesServer.start()
   private val baseJamesJmap: String = s"http://localhost:${server.mappedJmapPort}"
+  private val baseJamesWebsocket: String = s"ws://localhost:${server.mappedJmapPort}"
   lazy val protocolConf: Protocol = http.baseUrl(baseJamesJmap)
+    .wsBaseUrl(baseJamesWebsocket)
   private lazy val jamesJmap: JamesJmap = new JamesJmap(new URL(baseJamesJmap))
 
   protected def mappedJmapPort = server.mappedJmapPort
--- a/src/it/scala-2.12/org/apache/james/gatling/jmap/rfc8621/WebsocketSimpleIT.scala
+++ b/src/it/scala-2.12/org/apache/james/gatling/jmap/rfc8621/WebsocketSimpleIT.scala
@@ -1,0 +1,17 @@
+package org.apache.james.gatling.jmap.rfc8621
+
+import org.apache.james.gatling.Fixture
+import org.apache.james.gatling.jmap.draft.JmapIT
+import org.apache.james.gatling.jmap.rfc8621.scenari.WebsocketSimpleScenario
+
+import scala.concurrent.duration._
+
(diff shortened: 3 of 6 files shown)


src/main/scala-2.12/org/apache/james/gatling/jmap/rfc8621/scenari/WebsocketSimpleScenario.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import scala.concurrent.duration._
+
+class WebsocketSimpleScenario {
+  def generate(duration: Duration, userFeeder: UserFeederBuilder): ScenarioBuilder =
+    scenario("WebsocketSimpleScenario")
+      .feed(userFeeder)
+      .exec(SessionStep.retrieveAccountId)
+      .during(duration) {
+        exec(websocketConnect().onConnected(
```
chibenwa: Maybe add a delay before enabling the PUSH? (never know...)

chibenwa:
> It's a circle right in terms of user feeding? My memory is a bit rusty... 

For IT we don't care. Even a single user would do the JOB.

> So of course with only one user it opens the websocket first and tries to open new connections again (but failed because one opened already). So we can as well open multiple websocket connections per virtual user but it requires using the method wsName. 

Can you share the error? It is not normal, a single user is supposed to be able to open several WS on the James side?

> Do we need that, or should we just open one connection per user?

One connection per user seems good to me.

> The check fails. It tells me found nothing. The problem is I'm not sure what kind of answer I get. I might need to dig more on that. Could not find a way to try print the possible response we get for debugging too.

To be fair I do not care much about getting method responses over WS, I would care of having the stateChanges...

chibenwa:
```
[error] Uncaught exception when running org.apache.james.gatling.jmap.rfc8621.MailboxGetIT: java.lang.NoClassDefFoundError: io/netty/internal/tcnative/SSLPrivateKeyMethod

[error] sbt.ForkMain$ForkError: java.lang.NoClassDefFoundError: io/netty/internal/tcnative/SSLPrivateKeyMethod
```

https://travis-ci.com/github/linagora/james-gatling/builds/224140067

I saw the same things on @vttranlina laptop before leaving the office. We may need to upgrade gatling version...

Arsnael:
> Can you share the error? It is not normal, a single user is supposed to be able to open several WS on the James side?

That's not James being the problem, it's Gatling. I would need to use for each websocket call the `wsName` method where I give a unique name for each connection (to differentiate every connection from a user).

src/main/scala-2.12/org/apache/james/gatling/jmap/rfc8621/scenari/WebsocketSimpleScenario.scala
Arsnael: Nothing to do with it :)

Arsnael:
Ok thanks for the feedback and a good night rest... It's stupid actually (see the fixup). We can just yes open the websocket connection and then loop on stuff to do on it. Just should take that open connection out of the duration brackets... And about my check well the mailboxes were not provisioned. Now it works like a charm.

@chibenwa : I know you want to check the state instead on the push after modifying mailboxes, mails and stuff, and I will. I just started with something very simple to setup gatling with websockets and see if it's working. But from this now, should not be too hard to get there.

Result Locally:

```
================================================================================
2021-04-27 08:57:39                                          13s elapsed
---- Requests ------------------------------------------------------------------
> Global                                                   (OK=14     KO=0     )
> retrieveAccountId                                        (OK=1      KO=0     )
> getMailboxes                                             (OK=1      KO=0     )
> Open websocket                                           (OK=1      KO=0     )
> check Inbox                                              (OK=10     KO=0     )
> Close websocket                                          (OK=1      KO=0     )

---- WebsocketSimpleIT ---------------------------------------------------------
[##########################################################################]100%
          waiting: 0      / active: 0      / done: 1     
================================================================================

Simulation org.apache.james.gatling.jmap.rfc8621.WebsocketSimpleIT completed in 12 seconds
Parsing log file(s)...
Parsing log file(s) done
Generating reports...

================================================================================
---- Global Information --------------------------------------------------------
> request count                                         14 (OK=14     KO=0     )
> min response time                                     13 (OK=13     KO=-     )
> max response time                                    247 (OK=247    KO=-     )
> mean response time                                    78 (OK=78     KO=-     )
> std deviation                                         62 (OK=62     KO=-     )
> response time 50th percentile                         54 (OK=54     KO=-     )
> response time 75th percentile                         94 (OK=94     KO=-     )
> response time 95th percentile                        203 (OK=203    KO=-     )
> response time 99th percentile                        238 (OK=238    KO=-     )
> mean requests/sec                                  1.077 (OK=1.077  KO=-     )
---- Response Time Distribution ------------------------------------------------
> t < 800 ms                                            14 (100%)
> 800 ms < t < 1200 ms                                   0 (  0%)
> t > 1200 ms                                            0 (  0%)
> failed                                                 0 (  0%)
================================================================================

Reports generated in 0s.
Please open the following file: /home/rene/Dev/github/james-gatling/target/gatling-it/websocketsimpleit-20210427015725629/index.html
retrieveAccountId: percentage of failed requests is 0.0 : true
getMailboxes: percentage of failed requests is 0.0 : true
Open websocket: percentage of failed requests is 0.0 : true
check Inbox: percentage of failed requests is 0.0 : true
Close websocket: percentage of failed requests is 0.0 : true
[info] Simulation WebsocketSimpleIT successful.
[info] Simulation(s) execution ended.
[success] Total time: 42 s, completed Apr 27, 2021, 8:57:41 AM
```

Regarding the error from the CI, I will take a look :)

Arsnael:
> Regarding the error from the CI, I will take a look :)

Guilty: https://github.com/linagora/gatling-imap/pull/45

We might need to review all our deps one day here...

Arsnael:
For the record, I did not test yet the `WebsocketSimpleSimulation`. However the `WebsocketSimpleIT` is green on my local. I don't know if it is enough but let me know if it's not

src/main/scala-2.12/org/apache/james/gatling/jmap/rfc8621/scenari/WebsocketSimpleScenario.scala
```diff
@@ -0,0 +1,45 @@
...
+          ws.checkTextMessage("check response").check(jsonPath("$.methodResponses[0][1].created").find.saveAs("mailboxCreated"))))
+//        .await(2 seconds)(
+//          ws.checkTextMessage("check response").check(jsonPath("$").find.saveAs("mailboxCreated")))
+//        .await(2 seconds)(
+//          ws.checkTextMessage("check state").check(jsonPath("$").find.saveAs("mailboxState"))))
+//      .exec { session =>
+//        println(session("mailboxCreated").as[String])
+//        println(session("mailboxState").as[String])
+//        session
+//      }
```
chibenwa: Remove commented code

src/test/scala-2.12/org/apache/james/gatling/simulation/jmap/rfc8621/WebsocketSimpleSimulation.scala
```diff
@@ -0,0 +1,26 @@
+package org.apache.james.gatling.simulation.jmap.rfc8621
+
+import io.gatling.core.Predef._
+import io.gatling.core.scenario.Simulation
+import org.apache.james.gatling.control.{Password, User, UserFeeder, Username}
+import org.apache.james.gatling.jmap.rfc8621.scenari.WebsocketSimpleScenario
+import org.apache.james.gatling.simulation.{Configuration, HttpSettings}
+
+class WebsocketSimpleSimulation  extends Simulation {
```
chibenwa: IMO we do not even need the simulation now...

chibenwa (commented):
Looks really good to me. Koodo!

src/test/scala-2.12/org/apache/james/gatling/simulation/jmap/rfc8621/WebsocketSimpleSimulation.scala
chibenwa: IMO we do not even need the simulation now...

Arsnael: I have no issue of taking it out :)

src/main/scala-2.12/org/apache/james/gatling/jmap/rfc8621/scenari/WebsocketSimpleScenario.scala
Arsnael: oups, that's what I used for debug, will do

chibenwa (approved):

chibenwa:
Don't forget to reword to add ticket number though...

Arsnael:
> Don't forget to reword to add ticket number though...

We can when merging :)

Arsnael:
Reworded. The build was green btw

LanKhuat (approved):

PR #114, opened 2021-04-29
[FIX] Authentication management: better handle timeouts and missing v…

chibenwa:
…ariables

 - statusCodes are not defined upon timeouts which let to crashes while checking auth
 - Missing authentication tokens should trigger auth prior optimistic execution

 Undefined session variables led to scenarii crashes / bad interpretations. With
 the current fix, we avoid further errors upon timeouts and better handle missing
 session variables.

### DIFF
--- a/src/main/scala-2.12/org/apache/james/gatling/jmap/draft/RetryAuthentication.scala
+++ b/src/main/scala-2.12/org/apache/james/gatling/jmap/draft/RetryAuthentication.scala
@@ -9,10 +9,11 @@ import io.gatling.http.request.builder.HttpRequestBuilder.toActionBuilder
 import io.gatling.http.response.Response
 
 object RetryAuthentication {
+  val undefined = 900
 
   def execWithRetryAuthentication(scenario: HttpRequestBuilder, checks: Seq[HttpCheck]): ChainBuilder =
     doIf(session => !session.attributes.contains("accessTokenHeader")) {JmapAuthentication.authentication()}
-      .exec(session => session.set("statusCode", 900))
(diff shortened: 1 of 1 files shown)


src/main/scala-2.12/org/apache/james/gatling/jmap/draft/RetryAuthentication.scala
```diff
@@ -11,7 +11,9 @@ import io.gatling.http.response.Response
 object RetryAuthentication {
 
   def execWithRetryAuthentication(scenario: HttpRequestBuilder, checks: Seq[HttpCheck]): ChainBuilder =
-    exec(scenario.check(status.in(200, 401).saveAs("statusCode")).check(checkIfOk(checks): _*))
+    doIf(session => !session.attributes.contains("accessTokenHeader")) {JmapAuthentication.authentication()}
+      .exec(session => session.set("statusCode", 900))
```
mbaechler: what this `900` means?

chibenwa: Nothing, just specifying a random value here so that the check later don't complain upon timeouts that statusCode session variable is undefined...

Clearly it is a hack.

I had a quick time at it, and did not found better. However it did remove noise upon timeouts missing a status code in my runs... Which is nice!

mbaechler: What about, at least, giving this magic number a name and add a comment about it?

mbaechler (approved):

Arsnael (approved):

PR #125, opened 2021-06-10
Simulate ping for the push simulation

Arsnael:


src/main/scala-2.12/org/apache/james/gatling/jmap/rfc8621/scenari/PushPlatformValidationScenario.scala
```diff
@@ -72,16 +73,20 @@ class PushPlatformValidationScenario(minMessagesInMailbox: Int,
...
+                5.0 -> sendMessage(recipientFeeder),
+                30.0 -> openArbitrary.openArbitrary,
+                10.0 -> flagUpdate,
+                15.0 -> getNewState,
+                30.0 -> exec())
+                .asInstanceOf[ChainBuilder]
+                .pause(minWaitDelay, maxWaitDelay),
+
+              exec(echoPingWs)
+                .pause(pingInterval))
```
Arsnael: From what I understood, if I put multiple chains in an exec, they execute in parallel? But I'm not exactly sure... Guess we will see :)

chibenwa: No idea...

Arsnael: Conclusion: this was a terrible idea! Somehow it makes it opening much more websockets than the number of users running... Really odd. Trying to find other ways...

Arsnael:
It seemed to work on my local with 100 users running for 15 minutes. No websocket connection lost. Will test tomorrow on charge-01 on a bigger scale

Arsnael:
Jmap Push simulation with 3000 users, 15 mins injection, 45 mins run:

screenshot--2021 08 17-14_58_38

Getting some timeout on the ping response a few times but the websocket connection stayed alive until the end for all users :)

chibenwa (approved):

chibenwa:
(Though I might not be surprised that using JMAP methods for the ping might contribute to put some pressure on James ;-) )

Arsnael:
> (Though I might not be surprised that using JMAP methods for the ping might contribute to put some pressure on James ;-) )

Yeah totally agree on that, but it seems gatling does not manage the ping/pong mechanism at all regarding websockets so... couldn't find any better unfortunately^^'

chibenwa:
Nope nope that's fine, just that we should just be aware of it.

PR #132, opened 2022-03-17
Force pull of the latest tmail memory branch master image in Jenkinsfile

Arsnael:


quantranhong1999 (approved):

Jenkinsfile
```diff
@@ -8,6 +8,7 @@ pipeline {
         }
         stage('Test') {
             steps {
+                sh 'docker pull linagora/tmail-backend:memory-branch-master'
```
chibenwa, line 11: Do we need the same thing for gatling-imap?

chibenwa (approved):

Jenkinsfile
Arsnael: https://github.com/linagora/gatling-imap/pull/55

but as you approved I guess you saw it :)

PR #136, opened 2022-04-25
Another build for ISSUE-133

chibenwa:


Jenkinsfile
```diff
@@ -3,6 +3,7 @@ pipeline {
     stages {
         stage('Compile') {
             steps {
+                sh 'sbt reload'
```
Arsnael, line 6: Since the time I've been looking for this one... Cheers :)

chibenwa: Nope it don't work, it only reload existing project declarations without refreshing the clone...

Arsnael: Ah damn... tricky ^^'

PR #137, opened 2022-04-28
Auto-complete scenario should be more realistic

quantranhong1999:
Auto-complete requests were sent rapidly before which is too much in reality.

### DIFF
--- a/src/main/scala-2.12/org/apache/james/gatling/jmap/rfc8621/scenari/TmailContactAutocompleteScenario.scala
+++ b/src/main/scala-2.12/org/apache/james/gatling/jmap/rfc8621/scenari/TmailContactAutocompleteScenario.scala
@@ -24,12 +24,12 @@ class TmailContactAutocompleteScenario {
       .exec(SessionStep.retrieveAccountId)
       .exec(JmapMailbox.provisionUsersWithMessages(recipientFeeder, numberOfMessages = 5))
       .pause(2 seconds)
-      .randomSwitch(
-        70.0 -> exec((session: Session) => session.set(typeaheadKeyword, randomSubString(session.attributes(recipientSessionParam).asInstanceOf[String]))),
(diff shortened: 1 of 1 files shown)


src/main/scala-2.12/org/apache/james/gatling/jmap/rfc8621/scenari/TmailContactAutocompleteScenario.scala
```diff
@@ -24,12 +24,12 @@ class TmailContactAutocompleteScenario {
...
-        70.0 -> exec((session: Session) => session.set(typeaheadKeyword, randomSubString(session.attributes(recipientSessionParam).asInstanceOf[String]))),
-        30.0 -> exec((session: Session) => session.set(typeaheadKeyword, Random.alphanumeric.take(5).mkString(""))))
       .during(duration) {
-        exec(JmapTmailContact.getAutocomplete(typeaheadKeyword = typeaheadKeyword)
-          .check(statusOk, noError))
+        randomSwitch(
+          70.0 -> exec((session: Session) => session.set(typeaheadKeyword, randomSubString(session.attributes(recipientSessionParam).asInstanceOf[String]))),
+          30.0 -> exec((session: Session) => session.set(typeaheadKeyword, Random.alphanumeric.take(5).mkString(""))))
+          .exec(JmapTmailContact.getAutocomplete(typeaheadKeyword = typeaheadKeyword).check(statusOk, noError))
+          .pause(2 seconds)
```
Arsnael: I even personally think 2seconds is still too heavy... I would put it a bit higher like 5.

What people think on this?

quantranhong1999: Me I am Ok with 5.

PR #129, opened 2022-03-14
Update the latest Tmail image for Integration tests

quantranhong1999:
```
[error] Failed tests:
[error] 	org.apache.james.gatling.jmap.rfc8621.WebsocketSimpleIT
[error] 	org.apache.james.gatling.jmap.rfc8621.PushPlatformValidationIT
[error] (Gatling-it / test) sbt.TestsFailedException: Tests unsuccessful
```

Details: Only failing on Close websocket request
image

Me I do not think this is James' side. Finding something could be related to Gatling's side...

`REQUEST	1		Close websocket	1647249935294	1647249935311	KO	Premature close`

PR #139, opened 2022-05-06
Migrate gatling version to 3.4.2

Arsnael:


chibenwa:
```
[error] /home/jenkins/build/workspace/James_Gatling_build_PR-139/src/main/scala-2.12/org/apache/james/gatling/imap/scenari/ImapStoreScenario.scala:34:16: value sequence is not a member of object io.gatling.commons.validation.Validation
[error]     Validation.sequence(value.flags.map(flag => flag.el[String]).map(expr => expr(session)))
[error]                ^
[error] /home/jenkins/build/workspace/James_Gatling_build_PR-139/src/main/scala-2.12/org/apache/james/gatling/jmap/draft/JmapMessages.scala:146:15: object Json in package json cannot be accessed in package io.gatling.core.json
[error]             ${Json.stringify(queryParameters)}
[error]               ^
[error] /home/jenkins/build/workspace/James_Gatlig_build_PR-139/src/main/scala-2.12/org/apache/james/gatling/jmap/draft/JmapMessages.scala:206:29: not found: value Json
[error]             "properties": ${Json.stringify(properties)}
[error]                             ^
[error] /home/jenkins/build/workspace/James_Gatling_build_PR-139/src/main/scala-2.12/org/apache/james/gatling/jmap/draft/JmapMessages.scala:193:29: not found: value Json
[error]             "properties": ${Json.stringify(properties)}
[error]                             ^
[error] /home/jenkins/build/workspace/James_Gatling_build_PR-139/src/main/scala-2.12/org/apache/james/gatling/jmap/rfc8621/JmapEmail.scala:31:19: not found: value Json
[error]            |    ${Json.stringify(params)},
[error]                   ^
[error] /home/jenkins/build/workspace/James_Gatling_build_PR-139/src/main/scala-2.12/org/apache/james/gatling/jmap/rfc8621/JmapEmail.scala:72:35: not found: value Json
[error]            |      "properties": ${Json.stringify(properties)}
[error]                                   ^
[error] /home/jenkins/build/workspace/James_Gatling_build_PR-139/src/main/scala-2.12/org/apache/james/gatling/jmap/rfc8621/JmapEmail.scala:90:35: not found: value Json
[error]            |      "properties": ${Json.stringify(properties)}
[error]                                   ^
```

Arsnael:
Yes... I know :D 

`Validation.sequence` does not exist anymore... Could not find a solution yet

`Json` in Gatling core became private so can't access the method stringify anymore... thinking of doing like in james, we have the play libs, using the Json libs with implicits map conversions... or if there is a better idea :)

chibenwa:
For JSON do we really need a parser or string interpolation would be enough?

Arsnael:
Not sure how the string interpolation is interpreting maps?

chibenwa:
> Not sure how the string interpolation is interpreting maps?

Something as simple as...

```
properties.map(p => s"\"${p.getKey()}\":\"${p.getValue()}\"").collect(Collectors.joining(",");
```

Would probably do the job for a map that is flat.

(The scala syntax expressed here might not be the right ;-) )

Arsnael:
The problem is... some maps aren't flat

```
type JmapParameters = Map[String, Any]

[...]

def openpaasEmailQueryParameters(mailboxKey: String = "inboxID"): JmapParameters =
    Map("filter" -> Map("inMailbox" -> s"$${$mailboxKey}"),
      "comparator" -> List(Map("property" -> "sentAt", "isAscending" -> false)),
      "position" -> 0,
      "limit" -> 30)
```
is one example...

Arsnael:
But I'm wondering why we even bother with that honestly... Maybe a refactoring and we just balance the json string we need?

chibenwa:
Yes and it will avoid recomputing the JSON on each query...

Arsnael:
That should do it...

I still miss reactivating the close on websocket in concerned tests... However I tried quickly and it still fails locally on the IT test if I put it back

@quantranhong1999 are you sure the upgrade is enough? Didn't have time to dig more though

quantranhong1999:
> @quantranhong1999 are you sure the upgrade is enough? Didn't have time to dig more though

Sorry. I am not sure if this is 100% solving the failing tests. From what I did research, I just say that we could.

rf: https://github.com/linagora/james-gatling/pull/129#issuecomment-1066626849

chibenwa (approved):

vttranlina (approved):

chibenwa (approved):

PR #140, opened 2022-05-09
[CUSTOMER]  Massive operation simulation

quantranhong1999:


chibenwa (approved):
Please share a run result for this ;-)

src/main/scala-2.12/org/apache/james/gatling/imap/scenari/MassiveOperationScenario.scala
```diff
@@ -0,0 +1,78 @@
...
+  private val setAllMessagesInMailboxBDeletedFlagAndExpunge = exec(imap("storeAll").store(MessageRanges(From(1L)), StoreFlags.add(Silent.Enable(), "\\Deleted")).check(ok))
+    .exec(imap("expunge").expunge().check(ok))
+  private val deleteMailboxA_B_C = exec(imap("deleteFolder").deleteFolder(mailboxAName).check(ok))
+    .exec(imap("deleteFolder").deleteFolder(mailboxBName).check(ok))
+    .exec(imap("deleteFolder").deleteFolder(mailboxCName).check(ok))
+  private val createMailboxDWith1000SubMailboxes = exec(imap("createFolder").createFolder(mailboxDName).check(ok))
+    .exec(repeat(numberOfSubMailboxes, "loopId")(pause(gracePeriod).exec(exec(imap("createFolder").createFolder(s"$mailboxDName.$${loopId}").check(ok)))))
+  private val renameMailboxD = exec(imap("renameFolder").renameFolder(mailboxDName, mailboxDNewName).check(ok))
+  private val deleteMailboxD = exec(imap("deleteFolder").deleteFolder(mailboxDNewName).check(ok))
+    .exec(repeat(numberOfSubMailboxes, "loopId")(pause(gracePeriod).exec(exec(imap("deleteFolder").deleteFolder(s"$mailboxDNewName.$${loopId}").check(ok)))))
```
Arsnael, line 49: So it seems in IMAP if we delete a folder, it retroactively deletes all subfolders... (or so I think) so why do you need to try delete those subfolders after? those that work?

quantranhong1999: > those that work?

Deleting subfolders explicitly did work on James.

Arsnael: Ok... not sure to remember how it works then :D

chibenwa: When you don't know read the spec, Luke!

https://datatracker.ietf.org/doc/html/rfc3501#section-6.3.4

```
      The DELETE command MUST NOT remove inferior hierarchical names.
      For example, if a mailbox "foo" has an inferior "foo.bar"
      (assuming "." is the hierarchy delimiter character), removing
      "foo" MUST NOT remove "foo.bar".  It is an error to attempt to
      delete a name that has inferior hierarchical names and also has
      the \Noselect mailbox name attribute (see the description of the
      LIST response for more details).
  ```

PR #166, opened 2023-07-10
Support MAX_DURATION to force your run to terminate based on a durati…

vttranlina:
…on limit

- even if some virtual users are still running
- support get HTML result  when running hanging at 99%


chibenwa (approved):
VERY GOOD CATCH <3

Arsnael (approved):
Didn't know it was possible, very good contribution, definitely helpful!

PR #172, opened 2023-09-26
Add moving emails to JMAP benchmarks

chibenwa:
image

### DIFF
--- a/src/main/scala-2.13/org/apache/james/gatling/jmap/rfc8621/scenari/PushPlatformValidationScenario.scala
+++ b/src/main/scala-2.13/org/apache/james/gatling/jmap/rfc8621/scenari/PushPlatformValidationScenario.scala
@@ -5,10 +5,10 @@ import io.gatling.core.structure.{ChainBuilder, ScenarioBuilder}
 import io.gatling.http.Predef._
 import org.apache.james.gatling.control.RecipientFeeder.RecipientFeederBuilder
 import org.apache.james.gatling.control.UserFeeder.UserFeederBuilder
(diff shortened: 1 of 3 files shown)


src/main/scala-2.13/org/apache/james/gatling/jmap/rfc8621/scenari/PushPlatformValidationScenario.scala
```diff
@@ -18,16 +18,18 @@ object PushPlatformValidationScenario {
...
+  val emailId = "emailId"
   val messageIds = "messageIds"
   val randomMailbox = "randomMailbox"
 }
 
 class PushPlatformValidationScenario(minMessagesInMailbox: Int,
-                                     minWaitDelay: FiniteDuration = 20 seconds,
-                                     maxWaitDelay: FiniteDuration = 40 seconds) {
+                                     minWaitDelay: FiniteDuration = 2 seconds,
+                                     maxWaitDelay: FiniteDuration = 4 seconds) {
```
quantranhong1999: Q: We should revert the pause duration after the perf test MOVE, no? Otherwise, we would put much more stress on our JMAP simulation IMO.

Arsnael: Agreed, actually that change should not be in the code base :)

chibenwa:
```suggestion
                                     minWaitDelay: FiniteDuration = 20 seconds,
                                     maxWaitDelay: FiniteDuration = 40 seconds) {
```

Arsnael (approved):

PR #176, opened 2024-04-10
Fix Email/set patch update flags for JMAP RFC-8621

Arsnael:


quantranhong1999:
How are we checking if the response is OK? Just 200 OK?

Maybe we can do better e.g. check the response containing the `updated` keyword to prevent the same issues.

Arsnael:
> How are we checking if the response is OK? Just 200 OK?

> Maybe we can do better e.g. check the response containing the updated keyword to prevent the same issues.

Last fixup should address your concerns :)

quantranhong1999 (approved):

chibenwa (approved):

PR #178, opened 2024-04-29
Add simulation for full jmap without websockets

Arsnael:


src/main/scala-2.13/org/apache/james/gatling/jmap/rfc8621/scenari/JmapPlatformValidationScenario.scala
```diff
@@ -0,0 +1,82 @@
...
+    exec(JmapMailbox.getNewState(accountId, mailboxState))
+      .pause(1 second)
+      .exec(JmapEmail.getNewState(accountId, emailState))
+
+  val inboxHomeLoading: InboxLoadingScenario = new InboxLoadingScenario()
+  val openArbitrary: OpenEmailScenario = new OpenEmailScenario
+  val selectArbitrary: SelectMailboxScenario = new SelectMailboxScenario(minMessagesInMailbox)
+
+  def generate(duration: Duration, userFeeder: UserFeederBuilder, recipientFeeder: RecipientFeederBuilder): ScenarioBuilder =
+    scenario("PushPlatformValidation")
```
quantranhong1999: no more `Push` I guess?

quantranhong1999 (approved):

chibenwa (commented):
Can we get a sample run for this?

chibenwa (approved):

vttranlina (approved):

Arsnael:
> Can we get a sample run for this?

5K users:

jmap-nopush-5k

10k users:

jmap-nopush-10k

Not sure why the p99 for submitEmail is lower on 10k than 5k... Maybe unlucky? Also realized we probably didn't update for a while the gatling repo in our VM, I see a moveTwice operation that I don't have in last release perf tests for example. 

The perf for 5k seems a a bit worse than the ones with push. But wondering if it's because unlucky or because the repo was outdated for a while. I'm rerunning a 5k with the push to see.

Also 3 times less requests per second but that I think is because we don't need to ping for every request to keep the websocket active anymore

PR #180, opened 2024-09-09
Differentiate env configuration for scenario and injection duration o…

Arsnael:
…f a simulation

As we clearly dont necessary want the same duration for injection and scenario durations, having the same env variable for both makes no sense IMO.

### DIFF
--- a/src/test/scala-2.13/org/apache/james/gatling/simulation/Configuration.scala
+++ b/src/test/scala-2.13/org/apache/james/gatling/simulation/Configuration.scala
@@ -28,13 +28,23 @@ object Configuration {
     case _ => None
   }
 
+  val SCENARIO_DURATION_PROPERTY = Properties.envOrNone("SCENARIO_DURATION") match {
+    case Some(duration) => Some(duration.toInt minutes)
+    case _ => DURATION_PROPERTY
+  }
+
+  val INJECTION_DURATION_PROPERTY = Properties.envOrNone("INJECTION_DURATION") match {
+    case Some(duration) => Some(duration.toInt minutes)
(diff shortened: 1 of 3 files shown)


quantranhong1999 (approved):

src/test/scala-2.13/org/apache/james/gatling/simulation/Configuration.scala
```diff
@@ -23,7 +23,12 @@ object Configuration {
   val WEBADMIN_PROTOCOL = Properties.envOrElse("WEBADMIN_PROTOCOL", "http")
   val BaseJamesWebAdministrationUrl = new URI(s"$WEBADMIN_PROTOCOL://$WebadminServerHostName:$WEBADMIN_PORT").toURL
 
-  val DURATION_PROPERTY = Properties.envOrNone("DURATION") match {
+  val SCENARIO_DURATION_PROPERTY = Properties.envOrNone("SCENARIO_DURATION") match {
+    case Some(duration) => Some(duration.toInt minutes)
+    case _ => None
+  }
+
+  val INJECTION_DURATION_PROPERTY = Properties.envOrNone("INJECTION_DURATION") match {
```
chibenwa: We can avoid a breaking change by making both SCENARIO_DURATION and INJECTION_DURATION default to DURATION if unspecified.

quantranhong1999 (approved):

vttranlina (approved):

Arsnael:
Seems some IT tests failing but I see those same failures in latest master branch builds... will investigate

chibenwa (approved):

PR #184, opened 2025-01-21
ISSUE-183 JMAP scenario for triaging downloads

### LINKED ISSUE #183: JMAP scenario for triaging downloads
Run Email/query to filter out only messages with attachments

Save the attachments of up to the 20 first message with attachments into gatling session

Then randomly download an attachment on a periodic basis based on what is save onto the gatling session

DOD: run it on staging...

vttranlina:
resolve https://github.com/linagora/james-gatling/issues/183

### DIFF
--- a/sample.env
+++ b/sample.env
@@ -29,3 +29,8 @@
 # Integer value. Unit: Minutes. Force your run to terminate based on a duration limit. even if some virtual users are still running.
 # Default is 180 minutes
 MAX_DURATION=120
+
+# Boolean value. It provided the configure for DownloadAttachmentScenario.
+# Default is false
+# When it is true, it will create new emails with attachments before downloading them.
+#PROVISION_MAIL_WITH_ATTACHMENTS=true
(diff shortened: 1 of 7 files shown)


quantranhong1999 (approved):

Arsnael (approved):

chibenwa (approved):

chibenwa:
Can we also have a sample run for this?

vttranlina:
> Can we also have a sample run for this?

We have an issue with OpenSearch in the sandbox environment. 
Rene has asked the ops team to investigate. 
I will try again tomorrow. 
Is there a requirement for attachment size, such as 10MB...?

chibenwa:
> Is there a requirement for attachment size, such as 10MB...?

No no just a run here

Ideally 5 attachments of 1MB each per mail would be perfect ;-)

vttranlina:
The current commit support provision 1 attachment per mail.
Here is sample run on my local. (`PROVISION_MAIL_WITH_ATTACHMENTS=true`)

image

vttranlina:
A bit confuse when see the "Download Attachment" counter is very small than "email Get"
I rechecked code but it looks fine. 
Tried to replace `JmapHttp.download` by `JmapHttp.upload`, then the counter of "Upload" is ~ equal "email Get" (as expected)

Maybe "/download" is heavy (even when attachment is very very small) or other reason for `GET` method? any idea?

chibenwa:
@vttranlina the data is not representative

Email/get does json encoding / decoding too.

If you look at mean time dowloading an attachment is costlier than reading a mail...

Anyway this shall be enough for a differencial run.

PR #188, opened 2026-03-16
Jmap Email/Query: sort by receievdAt

Arsnael:
sentAt projections have been dropped in James and if not changed, we hit Opensearch instead of the email query view projection

## Release Notes

* **Bug Fixes**
  * Updated email sorting to use received date instead of sent date for more intuitive email organization across the application.


quantranhong1999 (approved):

chibenwa (commented):
G00d catch!

chibenwa (approved):

PR #189, opened 2026-03-27
Add email/query by keyword for jmap validation scenario

Arsnael:
I dont thnk a random word for keyword has much sense here. Opened to suggestions

## Summary by CodeRabbit

* **New Features**
  * Added an explicit keyword-based email filter and preserved existing text-based filtering; default sorting and pagination unchanged.
  * Submit-email now accepts an optional keyword (defaults to a random one).
  * Query operations can be invoked with an optional operation name for clearer tracing.

* **Tests**
  * Validation scenarios updated to exercise both text- and keyword-based queries and adjusted branch weights to include a dedicated keyword-query path.

### DIFF
--- a/src/main/scala-2.13/org/apache/james/gatling/jmap/rfc8621/JmapEmail.scala
+++ b/src/main/scala-2.13/org/apache/james/gatling/jmap/rfc8621/JmapEmail.scala
@@ -60,7 +60,7 @@ object JmapEmail {
        |""".stripMargin
   }
 
-  def filterKeywordQueryParameter(keyword: String = RandomStringGenerator.randomMeaningWord()): JmapParameters = {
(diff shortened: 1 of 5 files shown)


quantranhong1999:
> I dont thnk a random word for keyword has much sense here. Opened to suggestions

How about we have a list of keywords e.g. `keyword1`, `keyword2`... `keywordN`.

- Upon Email/set, we set those keywords randomly
- Upon Email/query hasKeyword, we randomly query one of those keywords?

Arsnael:
I mean yes but we have a massive existing dataset that does not follow this unfortunately^^'

Might see the benefits but later on

quantranhong1999:
> Might see the benefits but later on

I guess we can see the benefit after a few run(s).

Arsnael:
@quantranhong1999 let me know if this is fine with your suggestion

quantranhong1999 (approved):
Yes, it is. LGTM.

